PR #2, opened 2022-06-13
feat: Add end to end example

akocbek:
Issue: https://github.ibm.com/GoldenEye/issues/issues/2062

### DIFF
--- a/examples/version.tf
+++ b/examples/version.tf
@@ -6,7 +6,7 @@ terraform {
   required_providers {
     ibm = {
       source  = "IBM-Cloud/ibm"
-      version = ">= 1.41.1"
+      version = "1.41.0"
     }
   }
   required_version = ">= 1.0.0"


examples/main.tf
```diff
@@ -0,0 +1,21 @@
+##############################################################################
+# Resource Group
+##############################################################################
+
+resource "ibm_resource_group" "resource_group" {
+  name     = "${var.prefix}-rg"
+  quota_id = null
+}
```
ocofaigh, line 8: Can you update this example to support using an existing resource group

examples/variables.tf
```diff
@@ -0,0 +1,48 @@
...
+  default = "au-syd"
+}
+
+variable "prefix" {
+  description = "The prefix that you would like to append to your resources"
+  type        = string
+  default = "test-landing-zone-vpc"
+}
+
+variable "vpc_name" {
```
ocofaigh, line 27: no need for `var.vpc_name` and `var.prefix` in the example. Let just pass the `var.prefix` value for `vpc_name` and remove `var.vpc_name` from the example.

examples/version.tf
```diff
@@ -0,0 +1,16 @@
+##############################################################################
+# Terraform Providers
+##############################################################################
+
+terraform {
+  required_providers {
+    ibm = {
+      source  = "IBM-Cloud/ibm"
+      version = "~>1.41.0"
```
ocofaigh, line 9: Lock into lowest version in example to ensure we don't end up with invalid lower range. So that would be `1.41.0`

examples/version.tf
```diff
@@ -0,0 +1,16 @@
+##############################################################################
+# Terraform Providers
+##############################################################################
+
+terraform {
+  required_providers {
+    ibm = {
+      source  = "IBM-Cloud/ibm"
+      version = ">= 1.41.1"
```
ocofaigh, line 9: examples should lock into the lowest number in the range defined in the main modules version.tf. I see that is set to `~>1.41.0` meaning you should lock into `version = "1.41.0"` in the example

ocofaigh (approved):

PR #20, opened 2022-06-29
catalog onboarding information

argeiger:
### Description

This contains the files needed for onboarding this module into the IBM catalog


ocofaigh:
Hi @argeiger  - could you give some background on this? This module is already onboarded to the catalog. 
What happens with the data in `ibm_catalog.json` ? Is there something on the catalog side that parses this?
Also I see there is a version number hard coded. How do you plan to maintain this? New versions are released every time a PR is merged with the appropriate commit message (e.g. `fix` or `feat`). If the version bump is manual, then for sure we need to automate it before we merge this.

kbiegert:
@ocofaigh The ibm_catalog.json file is newer functionality for the catalog.  Previously the catalog used special sections in the README file to find stuff like compliance controls.  Now that and more are in the ibm_catalog.json file.  It is only consumed by the catalog during version import and has meta data that will populate areas of the catalog offering.

benbuchanan:
The version number can be excluded from the `ibm_catalog.json` file and the `--target-version` flag will be used if importing from the CLI.

ocofaigh:
Thanks @benbuchanan and @kbiegert - From a quick glance at internal script `catalog-onboard.sh`, it seems we are already using the `--target-version` flag. Since this is the case, should we proceed to remove the version from the json in this PR?
Is there any further changes needed in the catalog pipeline to pass this data to it? Or does it automatically pick it up when imported?

benbuchanan:
@ocofaigh You should be good to remove the version in this PR, and that should be the only change required.  Everything else will be picked up automatically on import.

ocofaigh:
@argeiger see https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/pull/20#issuecomment-1172434705

vburckhardt:
Looping back here - have we fully closed? The catalog json file is not in the main branch. I think we should go ahead, and use this, and then we can adjust when there is convergence towards using the unique metadata.yaml file.

ocofaigh:
Yep I think once we remove the version from the json we are good to merge

ocofaigh:
@kbiegert @argeiger Just checking back in on this one. Are we still in agreement that the version needs to be removed from the json in this PR and then we can merge?

kbiegert:
This is old and will be looked at later if any of these updates are needed.

PR #33, opened 2022-07-08
Add additional bug fixes  

argeiger:
### Description

Please include a summary of the change and which issue is fixed. Also, please include the motivation and context for the change. List any dependencies that are required for this change.

**Check the relevant boxes:**
- [ ] Bug fix (nonbreaking change that fixes an issue)
- [ ] New feature (nonbreaking change that adds functionality)
- [ X ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] CI related update (pipeline, etc.)

### Checklist

- [ ] If relevant, a test for the change has been added or updated as part of this PR.
- [ ] If relevant, documentation for the change has been added or updated as part of this PR.

### Merge

- Merge using "Squash and merge".
- Make sure to use a relevant [conventional commit](https://www.conventionalcommits.org/) message that is based on the PR contents. The commit message determines whether a new version of the modules needs to be released, and if so, which semver number to use).


variables.tf
```diff
@@ -2,6 +2,11 @@
 # Module Level Variables
 ##############################################################################
 
+variable "name" {
+  description = "Name for VPC"
```
ocofaigh, line 6: hmm, this might be confusing because this value will not actually be the vpc name, the vpc name will contain the prefix value too right?

variables.tf
```diff
@@ -2,6 +2,11 @@
 # Module Level Variables
 ##############################################################################
 
+variable "name" {
```
ocofaigh, line 5: what is the reasoning behind the variable name change? `vpc_name` -> `name` ? Just thinking can we avoid the breaking change here?

argeiger: We can but will be out of sync with what we have and will have to carry that forward with the other things that use it.  Not sure if there is a standard surrounding the `name` aspect but having `vpc` within is a little intuitive.  

Definitely understand the breaking aspect.

variables.tf
ocofaigh: hmm, but `prefix` is a required variable right now (it has no default value). Should we default it to null?

variables.tf
ocofaigh: As there is probably nobody consuming this module, I am OK with making this breaking change (and bumping the major version to 2.0.0) to get the module in sync with what is being used internally.

variables.tf
argeiger: Looking more into it, looks like `prefix` is used throughout without a check for null so we should make it required

ocofaigh (approved):

Issue #536, opened 2023-05-02
Enable creating network acls for disjoint ip address spaces


vburckhardt:
Context:
- network_cidr currently takes one string https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/blob/44fe07c4a2ae1c7b4110196595cd4625792ac692/variables.tf#L37
- this value is used to compute a set of network acl for ibm flows, and cross subnet flows. see https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/blob/main/network_acls.tf

Enhancement:
- case where the address space is disjoint (eg: multiple vpcs and on-prem direct link connected to same transit gateway) - it is not possible to have one single cidr covering all address space

Proposed solution:
- network_cidr to be renamed network_cidrs accepting a list of string (cidrs)
- iterate through each cidr in the logic at https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/blob/main/network_acls.tf#L6 https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/blob/main/network_acls.tf#L50 and https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/blob/main/network_acls.tf#L63

Aashiq-J:
Andrej suggested he is busy with another task and he hasn't started working on it. So, I have picked up this issue.

PR #542, opened 2023-05-23
feat: Enable creating network acls for disjoint ip address spaces

Aashiq-J:
### Description

case where the address space is disjoint (eg: multiple vpcs and on-prem direct link connected to same transit gateway) - it is not possible to have one single cidr covering all address space.
Loop through the list of all the network_cidr and create acl rules.

### Types of changes in this PR
- network_cidr to be renamed network_cidrs accepting a list of string (cidrs)
- iterate through each cidr in the logic at https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/blob/main/network_acls.tf#L6 https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/blob/main/network_acls.tf#L50 and https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/blob/main/network_acls.tf#L63

Issue : https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/issues/536

#### Changes that affect the core Terraform module or submodules
- [ ] Bug fix
- [x] New feature
- [ ] Dependency update

#### Changes that don't affect the core Terraform module or submodules
- [ ] Documentation update
- [ ] Other

#### Release required?
Identify the type of release. For information about the changes in a semantic versioning release, see [Release versioning](https://terraform-ibm-modules.github.io/documentation/#/versioning).

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [x] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

##### Release notes content

Ability to pass multiple network_cidrs and the network acls would be dynamically created according to network_cidrs.

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### Merge actions for mergers

- Use a relevant [conventional commit](https://www.conventionalcommits.org/) message that is based on the PR contents and any release notes provided by the PR author. The commit message determines whether a new version of the module is needed, and if so, which semver increment to use (major, minor, or patch).


network_acls.tf
```diff
@@ -39,40 +39,61 @@ locals {
...
-      destination = address.cidr
-      direction   = "outbound"
-      tcp         = null
-      udp         = null
-      icmp        = null
-    }
-  ]
+  ibm_cloud_internal_rules = flatten([
+    for rules in local.internal_rules : [
+      for index, cidrs in var.network_cidrs != null ? var.network_cidrs : ["0.0.0.0/0"] : {
```
akocbek, line 52: why do we override all properties defined already in `internal_rules`? why don't we add/override only `source`?

Aashiq-J: updated the logic.

dynamic_values.unit_tests.tf
```diff
@@ -66,7 +66,7 @@ locals {
   # tflint-ignore: terraform_unused_declarations
   assert_cluster_rule_exists_in_position_0 = regex("roks-create-worker-nodes-inbound", module.unit_tests.acl_map["acl"].rules[0].name)
   # tflint-ignore: terraform_unused_declarations
-  assert_cluster_rule_uses_network_cidr = regex("1.2.3.4/5", module.unit_tests.acl_map["acl"].rules[0].destination)
+  assert_cluster_rule_uses_network_cidr = regex("0.0.0.0/0", module.unit_tests.acl_map["acl"].rules[0].destination)
```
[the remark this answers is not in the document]
vburckhardt: @rajatagarwal-ibm 0.0.0.0/0 is closer to a real-world value than 1.2.3.4/5 as a cidr block.

vburckhardt:
This looks good. Thanks for the PR. Could you just put a screenshot of the network acl from the example where there are 2 values in the network_cidrs input.

dynamic_values/network_acls.tf
```diff
@@ -10,7 +10,7 @@ locals {
         [
           # These rules cannot be added in a conditional operator due to inconsistant typing
           # This will add all cluster_rules if the acl object contains prepend_ibm_rules as true
-          for rule in local.cluster_rules :
+          for rule in local.rules :
```
akocbek, line 13: is that change necessary? where local.rules is defined?

akocbek (changes requested):

dynamic_values/network_acls.tf
Aashiq-J: `cluster_rules` is the variable name for the acls with the static network_cidr and once the new acls are generated with the multiple cidrs the `cluster_rules` and the `app_rules `are concatenated into a single variable named `rules`

Aashiq-J:
@vburckhardt 
image
image

akocbek (approved):

vburckhardt:
@Aashiq-J @jojustin  - The allow all on IPs are not expected here. This would opened up too broadly the acls. Is this with the default configuration?

The network acls should be set as follows:

inbound:
each of the network_cidr -> each of the vpc address prefix

outbound:
 each of the vpc address prefix -> each of the network_cidr

jojustin:
Tx @vburckhardt for taking a look at it.  I merged as it was approved.  I should have checked the changes too.  Will take care of it going forward. 

@Aashiq-J can you please take a look at the above and fix it.

PR #581, opened 2023-07-21
fix: change the names of vpc_inbound_rule and vpc_outbound_rule

Aashiq-J:
### Description

Background: https://github.ibm.com/GoldenEye/issues/issues/4999#issuecomment-60308385
Modifying the current logic, where instead of using the index for cidr blocks alone, we could use indexes for both address_prefixes and cidr blocks. The modification can be done as follows:
```
   for address_index, address in data.ibm_is_vpc_address_prefixes.get_address_prefixes.address_prefixes : [
      for cidr_index, cidrs in var.network_cidrs != null ? var.network_cidrs : ["0.0.0.0/0"] :
      {
        name        = "ibmflow-allow-vpc-connectivity-inbound-${address_index}-${cidr_index}"
```

#### Changes that affect the core Terraform module or submodules
- [ ] Bug fix
- [ ] New feature
- [ ] Dependency update

#### Changes that don't affect the core Terraform module or submodules
- [ ] Documentation update
- [ ] Other

#### Release required?
Identify the type of release. For information about the changes in a semantic versioning release, see [Release versioning](https://terraform-ibm-modules.github.io/documentation/#/versioning).

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

##### Release notes content

If a release is required, replace this text with information that users need to know about the release. Write the release notes to help users understand the changes, and include information about how to update from the previous version.

Your notes help the merger write the commit message for the PR that is published in the release notes for the module.

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### Merge actions for mergers

- Use a relevant [conventional commit](https://www.conventionalcommits.org/) message that is based on the PR contents and any release notes provided by the PR author. The commit message determines whether a new version of the module is needed, and if so, which semver increment to use (major, minor, or patch).

### DIFF
--- a/network_acls.tf
+++ b/network_acls.tf
@@ -68,13 +68,13 @@ locals {
   ])
 
   vpc_inbound_rule = flatten([
-    for index, cidrs in var.network_cidrs != null ? var.network_cidrs : ["0.0.0.0/0"] : [
-      for address in data.ibm_is_vpc_address_prefixes.get_address_prefixes.address_prefixes :
+    for cidr_index, cidr in var.network_cidrs != null ? var.network_cidrs : ["0.0.0.0/0"] : [
+      for address_index, address in data.ibm_is_vpc_address_prefixes.get_address_prefixes.address_prefixes :
       {
-        name        = "ibmflow-allow-vpc-connectivity-inbound-${substr(address.id, -4, -1)}-${index}" # Providing unique rule names
+        name        = "ibmflow-allow-vpc-connectivity-inbound-${address_index}-${cidr_index}" # Providing unique rule names
         action      = "allow"
         source      = address.cidr
-        destination = cidrs
+        destination = cidr
         direction   = "inbound"
         tcp         = null
         udp         = null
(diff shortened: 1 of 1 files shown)


ocofaigh:
Is there any impact to consumers when renaming an ACL rule?

ocofaigh:
The test failed with `Resource(s) identified to be updated Name: network_acl Address: module.slz_vpc.ibm_is_network_acl.network_acl["vpc-acl"] Actions: [update]` which is expected due to the nature of the change here. But what I want to ensure if that there is no disruption to the customer when doing this update. It doesn't do a destroy and re-create, so at least that is a good sign.

Aashiq-J:
@ocofaigh ,
I tested the upgrade test on local while keeping a watch on the acls.
The acls rules are being deleted and re-created with the new name in the ibmcloud side. There is like a 5 sec window in which the rules are re-created. While, on the terraform side only the state files is being updated with the new names.

ocofaigh:
@Aashiq-J Thanks for checking. I'm not sure we should roll this out until we find a way to update ACLs non disruptively. There is an internal issue tracking that effort already https://github.ibm.com/GoldenEye/issues/issues/4517
Feel free to leave a comment in it with your observations.

Issue #598, opened 2023-08-11
Support passing null for prefix


ocofaigh:
When setting prefix to `null`, the following error is thrown:
```
│ Error: Invalid template interpolation value
│ 
│   on .terraform/modules/vpc/dynamic_values.tf line 7, in module "dynamic_values":
│    7:   prefix               = "${var.prefix}-${var.name}"
│     ├────────────────
│     │ var.prefix is null
│ 
│ The expression result is null. Cannot include a null value in a string template.
╵
╷
│ Error: Invalid template interpolation value
│ 
│   on .terraform/modules/vpc/main.tf line 94, in resource "ibm_is_public_gateway" "gateway":
│   94:   name           = "${var.prefix}-${var.name}-public-gateway-${each.key}"
│     ├────────────────
│     │ var.prefix is null
│ 
│ The expression result is null. Cannot include a null value in a string template.
```

It seems the code for the vpc name already supports prefix being null:
```
name                        = var.prefix != null ? "${var.prefix}-${var.name}-vpc" : "${var.name}-vpc"
```

However, the code for routing tables, flow logs, and public gateways are not. They should be updated to support a null prefix. Also ensure we update the prefix variable name to mention it supports passing `null` if you do not want to pass use a prefix
Make sure we test this end to end.

ocofaigh:
PR merged

PR #605, opened 2023-08-18
fix: add support for passing null in prefix

Khuzaima05:
### Description
Added  support for passing null in prefix

###issue : https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/issues/598

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [x] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### Merge actions for mergers


variables.tf
```diff
@@ -18,7 +18,7 @@ variable "region" {
 }
 
 variable "prefix" {
-  description = "The prefix that you would like to append to your resources"
+  description = "The prefix that you would like to append to your resources. It supports passing null"
```
ocofaigh, line 21: Maybe say this instead:

> Explicitly set to null if you do not wish to use a prefix.

examples/landing_zone/main.tf
```diff
@@ -68,7 +68,7 @@ module "management_vpc" {
 module "tg_gateway_connection" {
   source                    = "terraform-ibm-modules/transit-gateway/ibm"
   version                   = "2.1.1"
-  transit_gateway_name      = "${var.prefix}-tg"
+  transit_gateway_name      = var.prefix != null ? "${var.prefix}-tg" : "tg"
```
ocofaigh, line 71: I would not add this support to the example. The example should always use a prefix (so we avoid name clashes during test runs in our account)

examples/default/main.tf
```diff
@@ -19,19 +19,17 @@ data "ibm_resource_group" "existing_resource_group" {
...
   plan              = var.cos_plan
   location          = var.cos_location
 }
 
 resource "ibm_cos_bucket" "cos_bucket" {
-  count = var.enable_vpc_flow_logs ? 1 : 0
-
-  bucket_name          = "${var.prefix}-vpc-logs-cos-bucket"
+  count                = var.enable_vpc_flow_logs ? 1 : 0
+  bucket_name          = var.prefix != null ? "${var.prefix}-vpc-logs-cos-bucket" : "vpc-logs-cos-bucket"
```
ocofaigh, line 32: I would not add this support to the example. The example should always use a prefix (so we avoid name clashes during test runs in our account)

ocofaigh (changes requested):
@Khuzaima05 see comments. Can you also please paste a terraform plan output when prefix is set to null ?

Khuzaima05:
> @Khuzaima05 see comments. Can you also please paste a terraform plan output when prefix is set to null ?

Screenshot 2023-08-21 at 3 33 56 PM

ocofaigh:
So @Khuzaima05 from the screenshot (which looks like its a plan from an existing terraform state), the current code is giving an error. Are you looking in to that?

Khuzaima05:
> So @Khuzaima05 from the screenshot (which looks like its a plan from an existing terraform state), the current code is giving an error. Are you looking in to that?

This error got resolved.

Khuzaima05:
> > @Khuzaima05 see comments. Can you also please paste a terraform plan output when prefix is set to null ?
> 
> Screenshot 2023-08-21 at 3 33 56 PM

Recieving the same error in plan if we pass `NULL` in example.

examples/default/main.tf
```diff
@@ -19,19 +19,17 @@ data "ibm_resource_group" "existing_resource_group" {
 #############################################################################
 
 resource "ibm_resource_instance" "cos_instance" {
-  count = var.enable_vpc_flow_logs ? 1 : 0
-
-  name              = "${var.prefix}-vpc-logs-cos"
+  count             = var.enable_vpc_flow_logs ? 1 : 0
+  name              = var.prefix != null ? "${var.prefix}-vpc-logs-cos" : "vpc-logs-cos"
```
ocofaigh, line 23: @Khuzaima05 please remove this change. This does not have to support passing prefix as null. Infact prefix is required in the **example**

examples/default/main.tf
ocofaigh: @Khuzaima05 please remove this change. This does not have to support passing prefix as null. Infact prefix is required in the **example**

ocofaigh:
@Khuzaima05 Passing `null` for prefix does not work. I cloned your branch and updated the main.tf default example like this:
```
module "slz_vpc" {
  source                                 = "../../"
  resource_group_id                      = var.resource_group != null ? data.ibm_resource_group.existing_resource_group[0].id : ibm_resource_group.resource_group[0].id
  region                                 = var.region
  name                                   = var.name
  prefix                                 = null
  tags                                   = var.resource_tags
  access_tags                            = var.access_tags
  enable_vpc_flow_logs                   = var.enable_vpc_flow_logs
  create_authorization_policy_vpc_to_cos = var.create_authorization_policy_vpc_to_cos
  existing_cos_instance_guid             = ibm_resource_instance.cos_instance[0].guid
  existing_storage_bucket_name           = ibm_cos_bucket.cos_bucket[0].bucket_name
  address_prefixes                       = var.address_prefixes
  network_cidrs                          = var.network_cidrs
}
```

The error I get is:
```
│ Error: Invalid template interpolation value
│ 
│   on ../../dynamic_values.tf line 7, in module "dynamic_values":
│    7:   prefix               = "${var.prefix}-${var.name}"
│     ├────────────────
│     │ var.prefix is null
│ 
│ The expression result is null. Cannot include a null value in a string template.
```

ocofaigh:
Looks like there are 3 more places you need to make the change to support passing `null` for prefix:

- https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/blob/25cfdcaeedd851224c3a04c1492e722859307c30/dynamic_values.tf#L7
- https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/blob/25cfdcaeedd851224c3a04c1492e722859307c30/dynamic_values/main.tf#L20
- https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/blob/25cfdcaeedd851224c3a04c1492e722859307c30/network_acls.tf#L164

Please update these and show me the output of your test when passing `null` for prefix

ocofaigh:
@Khuzaima05 This is NOT how to test it. You do not want the pass `null` for the prefix variable in the example. you want you pass null into the VPC module directly. Just like I showed you in https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/pull/605#issuecomment-1689534168

Khuzaima05:
> 



> @Khuzaima05 This is NOT how to test it. You do not want the pass `null` for the prefix variable in the example. you want you pass null into the VPC module directly. Just like I showed you in [#605 (comment)](https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/pull/605#issuecomment-1689534168)

@ocofaigh , I tested accordingly, here is the terraform output (default example) :
Screenshot 2023-08-24 at 3 11 21 PM

ocofaigh:
Yes thanks, please push the code updates @Khuzaima05

ocofaigh (changes requested):
@Khuzaima05 your not supposed to commit the changes you made to test `null` in the examples. Thats only for local testing. Please revert them

Khuzaima05:
@ocofaigh Upgrade test is failing, Can I skip that.

Khuzaima05:
> @Khuzaima05 no - upgrade test should not fail with the changes being made in this PR - we need to figure out why its failing

@ocofaigh  Pipeline passed, Could you please review and merge.

ocofaigh (approved):

PR #620, opened 2023-09-06
test: added a new test to ensure null prefix works

kierramarie:
### Description

pr_test.go was updated to include a new test that ensures the prefix can be set to null. A new example was added to be used in this test. The example has the prefix hard-coded as null.

### Release required?

- [x] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### Merge actions for mergers

### DIFF
--- a/examples/no-prefix/README.md
+++ b/examples/no-prefix/README.md
@@ -1,0 +1,3 @@
+# Default Example
+
+An end to end example using the module's default variable values.
--- a/examples/no-prefix/README.md
+++ b/examples/no-prefix/README.md
@@ -1,3 +1,3 @@
-# Default Example
+# No Prefix Example
 
-An end to end example using the module's default variable values.
+An example similar to the basic example except in this example the prefix is hard-coded to be null. This example is used in a PR test.
--- a/examples/no-prefix/README.md
+++ b/examples/no-prefix/README.md
@@ -1,3 +1,3 @@
 # No Prefix Example
 
-An example similar to the basic example except in this example the prefix is hard-coded to be null. This example is used in a PR test.
+An example similar to the default example except in this example the prefix is hard-coded to be null.
--- a/examples/no-prefix/main.tf
+++ b/examples/no-prefix/main.tf
@@ -1,0 +1,56 @@
+##############################################################################
+# Resource Group
+# (if var.resource_group is null, create a new RG using var.prefix)
+##############################################################################
+
+resource "ibm_resource_group" "resource_group" {
+  count    = var.resource_group != null ? 0 : 1
+  name     = "${var.prefix}-rg"
+  quota_id = null
+}
+
+data "ibm_resource_group" "existing_resource_group" {
+  count = var.resource_group != null ? 1 : 0
+  name  = var.resource_group
+}
+
+#############################################################################
+# Provision cloud object storage and bucket
+#############################################################################
+
+resource "ibm_resource_instance" "cos_instance" {
+  count             = var.enable_vpc_flow_logs ? 1 : 0
+  name              = "${var.prefix}-vpc-logs-cos"
+  resource_group_id = var.resource_group != null ? data.ibm_resource_group.existing_resource_group[0].id : ibm_resource_group.resource_group[0].id
+  service           = "cloud-object-storage"
+  plan              = var.cos_plan
+  location          = var.cos_location
+}
+
+resource "ibm_cos_bucket" "cos_bucket" {
+  count                = var.enable_vpc_flow_logs ? 1 : 0
+  bucket_name          = "${var.prefix}-vpc-logs-cos-bucket"
+  resource_instance_id = ibm_resource_instance.cos_instance[0].id
+  region_location      = var.region
+  storage_class        = "standard"
+}
+
+#############################################################################
+# Provision VPC
+#############################################################################
+
+module "slz_vpc" {
+  source                                 = "../../"
+  resource_group_id                      = var.resource_group != null ? data.ibm_resource_group.existing_resource_group[0].id : ibm_resource_group.resource_group[0].id
+  region                                 = var.region
+  name                                   = var.name
+  prefix                                 = null
+  tags                                   = var.resource_tags
+  access_tags                            = var.access_tags
+  enable_vpc_flow_logs                   = var.enable_vpc_flow_logs
+  create_authorization_policy_vpc_to_cos = var.create_authorization_policy_vpc_to_cos
+  existing_cos_instance_guid             = ibm_resource_instance.cos_instance[0].guid
+  existing_storage_bucket_name           = ibm_cos_bucket.cos_bucket[0].bucket_name
+  address_prefixes                       = var.address_prefixes
+  network_cidrs                          = var.network_cidrs
+}
--- a/examples/no-prefix/main.tf
+++ b/examples/no-prefix/main.tf
@@ -15,37 +15,20 @@ data "ibm_resource_group" "existing_resource_group" {
 }
 
 #############################################################################
-# Locals
-#############################################################################
-locals {
-  access_tags                            = []
-  enable_vpc_flow_logs                   = true
-  cos_plan                               = "standard"
-  cos_location                           = "global"
-  create_authorization_policy_vpc_to_cos = true
-  network_cidrs                          = ["10.0.0.0/8", "164.0.0.0/8"]
-  address_prefixes = {
-    zone-1 = ["10.10.10.0/24"]
-    zone-2 = ["10.20.10.0/24"]
-    zone-3 = ["10.30.10.0/24"]
-  }
-}
-
-#############################################################################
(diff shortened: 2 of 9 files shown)


examples/no-prefix/README.md
```diff
@@ -0,0 +1,3 @@
+# Default Example
```
ocofaigh, line 1: update this title so its different from the actual default example

examples/no-prefix/README.md
```diff
@@ -0,0 +1,3 @@
...
+
+An end to end example using the module's default variable values.
```
ocofaigh, line 3: update description with more info on what this example shows

examples/no-prefix/variables.tf
```diff
@@ -0,0 +1,93 @@
...
+    error_message = "Keys for `use_public_gateways` must be in the order `zone-1`, `zone-2`, `zone-3`."
+    condition     = var.address_prefixes == null ? true : (keys(var.address_prefixes)[0] == "zone-1" && keys(var.address_prefixes)[1] == "zone-2" && keys(var.address_prefixes)[2] == "zone-3")
+  }
+}
+
+variable "network_cidrs" {
+  description = "List of Network CIDRs for the VPC. This is used to manage network ACL rules for cluster provisioning."
+  type        = list(string)
+  default     = ["10.0.0.0/8", "164.0.0.0/8"]
+}
```
ocofaigh, line 93: can we remove all these as variables and add them directly into the main.tf. It makes it easier for consumers to see what kind of values need to be passed for different vars, but it also prevents people changing variable values which may break the example.
(This is something we need to roll out to all repo examples eventually)

examples/no-prefix/README.md
```diff
@@ -1,3 +1,3 @@
-# Default Example
+# No Prefix Example
 
-An end to end example using the module's default variable values.
+An example similar to the basic example except in this example the prefix is hard-coded to be null. This example is used in a PR test.
```
ocofaigh, line 3: there is no basic example - its called default. Remove the line `This example is used in a PR test.`

examples/no-prefix/variables.tf
ocofaigh: @kierramarie still outstanding

kierramarie: which variables do you want to remain in the variables.tf?

tests/pr_test.go
```diff
@@ -54,6 +57,29 @@ func TestRunBasicExample(t *testing.T) {
 	assert.NotNil(t, output, "Expected some output")
 }
 
+func TestRunNull(t *testing.T) {
```
toddgiguere, line 60: I think we should change this test name to `TestRunNoPrefixExample`

examples/no-prefix/main.tf
```diff
@@ -0,0 +1,56 @@
...
+
+#############################################################################
+# Provision VPC
+#############################################################################
+
+module "slz_vpc" {
+  source                                 = "../../"
+  resource_group_id                      = var.resource_group != null ? data.ibm_resource_group.existing_resource_group[0].id : ibm_resource_group.resource_group[0].id
+  region                                 = var.region
+  name                                   = var.name
```
ocofaigh, line 46: don't we now want to update this to `name = var.prefix` so our tests will pass a unique prefix to it?

toddgiguere (changes requested):

examples/no-prefix/main.tf
ocofaigh: ah actually never mind, we are doing this directly in the test

examples/no-prefix/variables.tf
```diff
@@ -0,0 +1,93 @@
...
+
+variable "region" {
+  description = "The region to which to deploy the VPC"
+  type        = string
+  default     = "us-south"
+}
+
+variable "prefix" {
+  description = "The prefix that you would like to append to your resources"
```
ocofaigh, line 14: well this is not actually true - its used for all resources except the ones created in the VPC module itself.

examples/no-prefix/main.tf
kierramarie: The prefix should be null. We want to pass a unique name to it.

toddgiguere: @ocofaigh in the pr_test code there is a different unique value being generated for the name, and it is passed into the name variable for the test and checked afterwards. The prefix and name will use different and random/unique values.

Done this way because the unit test doesn't have direct access to the unique value used as part of the prefix, and the entire prefix is already a bit long.

examples/no-prefix/variables.tf
toddgiguere: and `name`, we still need the name input for this particular example (see other comment about how we are using that input in pr_test)

examples/no-prefix/main.tf
```diff
@@ -14,21 +14,38 @@ data "ibm_resource_group" "existing_resource_group" {
...
   service           = "cloud-object-storage"
-  plan              = var.cos_plan
-  location          = var.cos_location
+  plan              = local.cos_plan
+  location          = local.cos_location
 }
 
 resource "ibm_cos_bucket" "cos_bucket" {
-  count                = var.enable_vpc_flow_logs ? 1 : 0
+  count                = local.enable_vpc_flow_logs ? 1 : 0
```
ocofaigh, line 48: no need for a count now that we are hard coding `enable_vpc_flow_logs` to `true`

examples/no-prefix/main.tf
```diff
@@ -0,0 +1,60 @@
...
+  count = var.resource_group != null ? 1 : 0
+  name  = var.resource_group
+}
+
+#############################################################################
+# Provision cloud object storage and bucket
+#############################################################################
+
+resource "ibm_resource_instance" "cos_instance" {
+  count             = 1
```
ocofaigh, line 22: you dont need this count

examples/no-prefix/main.tf
```diff
@@ -0,0 +1,60 @@
...
+  name              = "${var.prefix}-vpc-logs-cos"
+  resource_group_id = var.resource_group != null ? data.ibm_resource_group.existing_resource_group[0].id : ibm_resource_group.resource_group[0].id
+  service           = "cloud-object-storage"
+  plan              = "standard"
+  location          = "global"
+}
+
+resource "ibm_cos_bucket" "cos_bucket" {
+  count                = 1
```
ocofaigh, line 31: you dont need this count

examples/no-prefix/main.tf
```diff
@@ -0,0 +1,60 @@
...
+  source                                 = "../../"
+  resource_group_id                      = var.resource_group != null ? data.ibm_resource_group.existing_resource_group[0].id : ibm_resource_group.resource_group[0].id
+  region                                 = var.region
+  name                                   = var.name
+  prefix                                 = null
+  tags                                   = var.resource_tags
+  access_tags                            = []
+  enable_vpc_flow_logs                   = true
+  create_authorization_policy_vpc_to_cos = true
+  existing_cos_instance_guid             = ibm_resource_instance.cos_instance[0].guid
```
ocofaigh, line 52: after removing the count, this can be updated to `ibm_resource_instance.cos_instance.guid`

examples/no-prefix/main.tf
```diff
@@ -0,0 +1,60 @@
...
+  existing_storage_bucket_name           = ibm_cos_bucket.cos_bucket[0].bucket_name
```
ocofaigh, line 53: after removing the count, this can be updated to `ibm_cos_bucket.cos_bucket.bucket_name`

ocofaigh (changes requested):
see comments on how to remove count

ocofaigh (changes requested):
Looks like we need to be more specific with our unique prefxi generation. It can't start with a number:

```
TestRunNoPrefixExample 2023-09-07T09:11:36Z logger.go:66: │ Error: "name" ("9cyinv-vpc") should match regexp ^([a-z]|[a-z][-a-z0-9]*[a-z0-9])$ 
TestRunNoPrefixExample 2023-09-07T09:11:36Z logger.go:66: │ 
TestRunNoPrefixExample 2023-09-07T09:11:36Z logger.go:66: │   with module.slz_vpc.ibm_is_vpc.vpc,
TestRunNoPrefixExample 2023-09-07T09:11:36Z logger.go:66: │   on ../../main.tf line 6, in resource "ibm_is_vpc" "vpc":
TestRunNoPrefixExample 2023-09-07T09:11:36Z logger.go:66: │    6:   name                        = var.prefix != null ? "${var.prefix}-${var.name}-vpc" : "${var.name}-vpc"
```

toddgiguere:
@kierramarie We can add a small prefix string to it, to ensure it starts with a letter.

toddgiguere (approved):

PR #621, opened 2023-09-07
fix: convert default sg/acl rule cleaning to provider option

toddgiguere:
### Description

Use new VPC resource provider option to clean rules from default VPC security group and ACL, retire the workaround scripts.

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [x] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

##### Release notes content

IBM Cloud Terraform provider v1.56.0 has added a new option for the "ibm_is_vpc" resource that will cause the default VPC ACL and Security Group to contain no rules (empty).

This release will retire the usage of existing backend scripts to remove all rules from the VPC default ACL and SG, and instead use this new provider option to accomplish the same feature.

Input variable changes:
The new provider option covers both security group and ACL in one variable, so we will be deprecating the individual "clean_" variables and replace them with a new single boolean to enable the feature.
- REMOVED: `clean_default_security_group` and `clean_default_acl`
- ADDED: `clean_default_sg_acl`, if set to "true" will trigger new VPC option to have empty default groups (default is "false")
- CHANGED: `security_group_rules` now has a default of empty, instead of a broad default inbound rule that may not be desired
- Validation has been added to make sure that `clean_default_sg_acl` has not been set to "true" while having rules specified in the `security_group_rules` input, which are in direct conflict with each other

Upgrade Notes:
If you have already deployed module with the "clean_" variables not specified, you should see no difference after upgrade.

If you have already deployed module with "clean_" variables set to "true/false", you will get an error after upgrade due to those variables being removed, and you should set the new `clean_default_sg_acl` variable instead. During a plan phase after upgrading, you may see the following resources marked for DESTROY, this is expected as these are the retired scripts that handled this feature in the past, and have been removed:
module.slz_vpc.null_resource.clean_default_acl[0] will be destroyed
module.slz_vpc.null_resource.clean_default_security_group[0] will be destroyed

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### Merge actions for mergers


ocofaigh (commented):
@toddgiguere What are your thoughts on just removing the old vars all together and including it in the release notes? That way we don't have to come back with a follow up PR to remove them in the future. Either way, this will be a minor version release

toddgiguere:
Latest commit has the deprecated variables removed entirely, the default for `security_group_rules` has been set to empty array, and a new validation has been added to check if consumer has set the security group options in conflicting way.

toddgiguere:
As anticipated, the UPGRADE test has failed due to the removal of the default security group rule that has in the past been added due to the default value of the input variable `security_group_rules`. Now that the default value has been removed and replaced with empty array, on upgrade the following resource is marked for destroy:

Resource(s) identified to be destroyed Name: default_vpc_rule Address: module.slz_vpc.ibm_is_security_group_rule.default_vpc_rule["default-sgr"] Actions: [delete]

This is the only resource marked by the upgrade test.

ocofaigh (approved):

Issue #599, opened 2023-08-11
Need a support for existing VPC functionality


digvijay-ukirde:
### Description

Customers usually have an existing VPC and expect the solution (VSIs) to be deployed in the same.
Requirements:

1. **Existing VPC**: If the customer provides a VPC name, we should not create VPC and address prefix (default=null); and use the VPC to deploy the resources (subnet, VSI, SG etc)
2. **Existing Subnet**: If customer provide a subnet name, we should skip Subnet creation (default=null) and use the subnet to deploy the resources (VSI, SG, Gateway etc)
3. **Existing Gateway**: If customer provide a VPC where gateway already exists for a particular zone then use it and skip creating gateway

### New or affected modules

---

By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/terraform-ibm-modules/documentation/blob/main/CODE_OF_CONDUCT.md)

vburckhardt:
- On VPC: current thinking is to see whether we could enhance landing zone module to skip vpc creation, in other word empty array in vpcs field.
   - scenario of existing vpc - enable creation of subnets / skip creation of subnets

- Expose the vsi secondary interface inputs in the landing zone module https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/blob/02648ac36c2eb15d9bceb8c43f200c6418a38735/variables.tf#L270

Aayush-Abhyarthi:
Currently looking for ways to use skip the vpc creation in the module so that the passed vpc can be used .

ocofaigh:
This support was added in https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/releases/tag/v7.12.0
We are now looking to see if we can add support to the SLZ solution. Tracked at https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/issues/635

PR #638, opened 2023-10-11
feat: add existing vpc functionality

Aayush-Abhyarthi:
### Description

https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/issues/599

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [x] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers

### DIFF
--- a/examples/basic/main.tf
+++ b/examples/basic/main.tf
@@ -4,7 +4,7 @@
 
 module "resource_group" {
   source  = "terraform-ibm-modules/resource-group/ibm"
-  version = "1.1.0"
+  version = "1.1.4"
   # if an existing resource group is not set (null) create a new one using prefix
   resource_group_name          = var.resource_group == null ? "${var.prefix}-resource-group" : null
   existing_resource_group_name = var.resource_group
--- a/examples/default/main.tf
+++ b/examples/default/main.tf
@@ -4,7 +4,7 @@
 
 module "resource_group" {
   source  = "terraform-ibm-modules/resource-group/ibm"
-  version = "1.1.0"
+  version = "1.1.4"
   # if an existing resource group is not set (null) create a new one using prefix
   resource_group_name          = var.resource_group == null ? "${var.prefix}-resource-group" : null
   existing_resource_group_name = var.resource_group
--- a/examples/hub-spoke-delegated-resolver/main.tf
+++ b/examples/hub-spoke-delegated-resolver/main.tf
@@ -4,7 +4,7 @@
 
 module "resource_group" {
   source  = "terraform-ibm-modules/resource-group/ibm"
-  version = "1.1.0"
+  version = "1.1.4"
   # if an existing resource group is not set (null) create a new one using prefix
   resource_group_name          = var.resource_group == null ? "${var.prefix}-resource-group" : null
   existing_resource_group_name = var.resource_group
--- a/examples/hub-spoke-manual-resolver/main.tf
+++ b/examples/hub-spoke-manual-resolver/main.tf
@@ -4,7 +4,7 @@
 
 module "resource_group" {
   source  = "terraform-ibm-modules/resource-group/ibm"
-  version = "1.1.0"
+  version = "1.1.4"
   # if an existing resource group is not set (null) create a new one using prefix
   resource_group_name          = var.resource_group == null ? "${var.prefix}-resource-group" : null
   existing_resource_group_name = var.resource_group
--- a/examples/landing_zone/main.tf
+++ b/examples/landing_zone/main.tf
@@ -4,7 +4,7 @@
 
 module "resource_group" {
   source  = "terraform-ibm-modules/resource-group/ibm"
-  version = "1.1.0"
+  version = "1.1.4"
   # if an existing resource group is not set (null) create a new one using prefix
   resource_group_name          = var.resource_group == null ? "${var.prefix}-resource-group" : null
   existing_resource_group_name = var.resource_group
@@ -17,7 +17,7 @@ module "resource_group" {
 module "cos_bucket" {
   count                  = var.enable_vpc_flow_logs ? 1 : 0
   source                 = "terraform-ibm-modules/cos/ibm"
-  version                = "7.0.3"
+  version                = "7.0.5"
   resource_group_id      = module.resource_group.resource_group_id
   region                 = var.region
   cross_region_location  = null
--- a/examples/no-prefix/main.tf
+++ b/examples/no-prefix/main.tf
@@ -4,7 +4,7 @@
 
 module "resource_group" {
   source  = "terraform-ibm-modules/resource-group/ibm"
-  version = "1.1.0"
+  version = "1.1.4"
   # if an existing resource group is not set (null) create a new one using prefix
   resource_group_name          = var.resource_group == null ? "${var.prefix}-resource-group" : null
   existing_resource_group_name = var.resource_group


examples/default/variables.tf
```diff
@@ -16,6 +16,11 @@ variable "prefix" {
   default     = "test-landing-zone"
 }
 
+variable "is_vpc_existing" {
+  description = "Indicates whether user wants to use an existing vpc or create a new one."
+  type        = bool
+}
+
```
shemau, line 23: Based partially on the COS module as example.

This support almost certainly requires two new properties (to avoid the terraform feature about using values created at apply time in counts). It likely also requires an addition example to validate using existing values.

For variables to be consistent with the COS module (there may be other conflicting examples you would prefer to follow)

existing_cos_instance_id and create_cos_instance become existing_vpc_id (string) and create_vpc (bool), extending to existing_subnet_ids (map(string)) and create_subnets (bool)... etc.

main.tf
```diff
@@ -24,6 +25,11 @@ resource "ibm_is_vpc" "vpc" {
   no_sg_acl_rules             = var.clean_default_sg_acl
 }
 
+data "ibm_is_vpc" "vpc" {
```
shemau, line 28: A local will be required to determine which ibm_is_vpc.vpc is required

```hcl
locals {
  vpc_id = var.create_vpc ? var.existing_vpc_id : ibm_is_vpc.vpc.id
}
```

IF it is only the ID that is used, there may be no need to data resource lookup.

Additional validation is required to detect errors early (plan), so make sure that if create_vpc is false, then existing_vp_id MUST be passed.

shemau (dismissed):
The landing-zone-vsi module already ONLY works on an existing VPC and existing subnets.

These feels like it might not be the right place for this code, it seems that calling a module that exists to create a VPC with an existing VPC is wrong.

Discuss at a deep dive, but maybe the features are required in one or more of the profiles or root module in landing-zone to avoid this module being called in the first place?

ocofaigh:
@Aayush-Abhyarthi We had a call today about this. What we are going to try implement is the ability to pass an **existing VPC** to the module, and alternatively the ability to pass an **existing VPC and existing subnets**.
We should add an example to test this use case, but one thing you need to update in the code is the outputs. Ensure that all of the outputs work when using existing VPC and subnets. Currently the outputs will only work when creating new VPC.

variables.tf
```diff
@@ -3,10 +3,22 @@
 ##############################################################################
 
 variable "name" {
-  description = "Name for VPC"
+  description = "Name for VPC."
   type        = string
 }
 
+variable "create_vpc" {
```
daniel-butler-irl, line 10: I don't think we need `create_vpc` and `existing_vpc_id `. Just do if `existing_vpc_id ` is null create a new vpc. @ocofaigh do you agree?

ocofaigh: We do need it - we will face issues where if you install the VPC in same terraform apply, this will fail saying it can't compute the plan on an unknown value. so hence it needs a boolean (we found this out the hard way before)

examples/existing_vpc/README.md
```diff
@@ -0,0 +1,3 @@
+# Basic Example
```
ocofaigh, line 1: There is already a "Basic Example". I would rename this to "Existing VPC and subnets example"

examples/existing_vpc/main.tf
```diff
@@ -0,0 +1,58 @@
...
+module "slz_vpc" {
+  source            = "../../"
+  resource_group_id = module.resource_group.resource_group_id
+  region            = var.region
+  tags              = var.resource_tags
+  create_vpc        = false
+  existing_vpc_id   = resource.ibm_is_vpc.example_vpc.id
+  create_subnets    = false
+
+  subnets = {
```
ocofaigh, line 32: You have set `create_subnets` to `false` here, so I expect to see `existing_subnet_ids` to be used, so you will also need to create subnets in the "example_vpc" VPC before calling this module.

ocofaigh (changes requested):
@Aayush-Abhyarthi I think your updates to the outputs are wrong. You changed `vpc_name` and `vpc_crn` outputs to now use the `data.ibm_is_vpc` data lookup, however, this data lookup has a count on it that means it only executes if `create_vpc = false`. In order to ensure there are no regressions introduced here with the outputs, I think you need to add ALL outputs to all examples that the tests run, and add logic to all the tests to ensure each of the outputs has a value.

variables.tf
```diff
@@ -3,8 +3,21 @@
 ##############################################################################
 
 variable "name" {
-  description = "Name for VPC"
+  description = "Name for VPC."
```
ocofaigh, line 6: Perhaps update to: "The name to give the newly provisioned VPC. Only used if 'create_vpc' is true."

variables.tf
```diff
@@ -3,8 +3,21 @@
...
   type        = string
+  default     = null
```
ocofaigh, line 8: hmm, by setting the default value to null, it means the current set of the module variable default values won't work out of the box (since `create_vpc` is `true` by default, meaning `name` is required). Perhaps we default the value to some real string like "dev" ?

variables.tf
```diff
@@ -20,6 +33,7 @@ variable "region" {
 variable "prefix" {
   description = "The prefix that you would like to append to your resources. Explicitly set to null if you do not wish to use a prefix."
```
ocofaigh, line 34: Just noticed the word "append" here would indicate the value is added to the end of the resource name, which is not true. Suggest rewording to: "The value that you would like to prefix to the name of the resources provisioned by this module. Explicitly set to null if you do not wish to use a prefix."

.ipynb_checkpoints/Untitled-checkpoint.ipynb
```diff
@@ -0,0 +1,6 @@
+{
+ "cells": [],
+ "metadata": {},
+ "nbformat": 4,
+ "nbformat_minor": 5
+}
```
ocofaigh, line 6: what is the file, was it accidentally committed?

Untitled.ipynb
```diff
@@ -0,0 +1,98 @@
+{
```
ocofaigh, line 1: what is the file, was it accidentally committed?

default_security_group.tf
```diff
@@ -12,7 +12,7 @@ locals {
 
 resource "ibm_is_security_group_rule" "default_vpc_rule" {
   for_each  = local.security_group_rule_object
-  group     = ibm_is_vpc.vpc.default_security_group
+  group     = ibm_is_vpc.vpc[0].default_security_group
```
ocofaigh, line 15: Hmm, this means this code block will only occur if a new VPC is created. Won't we want it to happen on an existing VPC too if one is passed ? @toddgiguere @digvijay-ukirde ?

examples/existing_vpc/README.md
```diff
@@ -0,0 +1,3 @@
+# Existing VPC and subnets Example
+
+A very basic example to provision SLZ VPC with existing vpc.
```
ocofaigh, line 3: It's not really a basic example. It's an example of calling the module with an existing VPC and subnets

examples/existing_vpc/main.tf
```diff
@@ -0,0 +1,68 @@
...
+module "resource_group" {
+  source  = "terraform-ibm-modules/resource-group/ibm"
+  version = "1.1.0"
+  # if an existing resource group is not set (null) create a new one using prefix
+  resource_group_name          = var.resource_group == null ? "${var.prefix}-resource-group" : null
+  existing_resource_group_name = var.resource_group
+}
+
+#############################################################################
+# Provision VPC
```
ocofaigh, line 14: It probably doesn't make sense to provision the VPC and subnets in the example here, as this is not something someone using existing vpc + subnets resources would be doing. We could maybe have terraform code in a different folder that provisions the VPC + subnets, and have our test call that code first, and pass the VPC + subnet details to this example. We did this in the [TestRunExistingGroupExample](https://github.com/terraform-ibm-modules/terraform-ibm-resource-group/blob/0058f23ed3bab77970a20d5791d4bb705fc28a95/tests/pr_test.go#L41C6-L41C33) test if you want to use it for reference. 
Suggest getting the rest of the code up to date an working first before updating this though.

examples/existing_vpc/main.tf
```diff
@@ -0,0 +1,68 @@
...
+
+module "slz_vpc" {
+  source              = "../../"
+  resource_group_id   = module.resource_group.resource_group_id
+  region              = var.region
+  tags                = var.resource_tags
+  create_vpc          = false
+  existing_vpc_id     = resource.ibm_is_vpc.example_vpc.id
+  create_subnets      = false
+  existing_subnet_ids = rersource.ibm_is_subnet.testacc_subnet.id
```
ocofaigh, line 39: how does this make sense? The `existing_subnet_ids` is a map value, yet you are passing a single string to it?

examples/existing_vpc/main.tf
```diff
@@ -0,0 +1,68 @@
...
+
+
+  subnets = {
```
ocofaigh, line 42: I don't see this being used anywhere?

examples/existing_vpc/output.tf
```diff
@@ -0,0 +1,13 @@
...
+
+output "vpc_id" {
+  value       = module.slz_vpc.vpc_id
+  description = "VPC id"
+}
+
+output "vpc_crn" {
+  value       = module.slz_vpc.vpc_crn
+  description = "VPC crn"
+}
```
ocofaigh, line 13: In order to ensure we have the outputs correct for this use case, can you output everything that the root level module outputs so we can check all values in the test. Thanks

examples/existing_vpc/variables.tf
```diff
@@ -0,0 +1,35 @@
...
+variable "region" {
+  description = "The region to which to deploy the VPC"
+  type        = string
+  default     = "us-south"
+}
+
+variable "prefix" {
+  description = "The prefix that you would like to append to your resources"
+  type        = string
+  default     = "basic-slz-vpc"
```
ocofaigh, line 16: `basic-slz-vpc` -> `existing-slz-vpc`

examples/existing_vpc/variables.tf
```diff
@@ -0,0 +1,35 @@
...
+}
+
+variable "name" {
+  description = "The name of the vpc"
+  type        = string
+  default     = "vpc"
+}
```
ocofaigh, line 23: Name won't be needed if using an existing VPC

main.tf
```diff
@@ -59,6 +66,15 @@ resource "ibm_is_vpc_dns_resolution_binding" "vpc_dns_resolution_binding_crn" {
   }
 }
 
+data "ibm_is_vpc" "vpc" {
+  count = var.create_vpc == false ? 1 : 0
+  id    = var.existing_vpc_id
```
ocofaigh, line 71: I think you will need to specify the resource group here, meaning you need a new inout variable called `existing_resource_group_id`

outputs.tf
```diff
@@ -4,17 +4,17 @@
 
 output "vpc_name" {
   description = "Name of VPC created"
-  value       = ibm_is_vpc.vpc.name
+  value       = var.create_vpc == true ? var.name : data.ibm_is_vpc.vpc[0].name
```
ocofaigh, line 7: You should not use variables in outputs like this. Always pull a value from the resource itself. So change `var.name` -> `ibm_is_vpc[0].vpc.name`

outputs.tf
```diff
@@ -4,17 +4,17 @@
...
 output "vpc_id" {
   description = "ID of VPC created"
-  value       = ibm_is_vpc.vpc.id
+  value       = local.vpc_id
 }
 
 output "vpc_crn" {
   description = "CRN of VPC created"
-  value       = ibm_is_vpc.vpc.crn
+  value       = var.create_vpc == true ? resource.ibm_is_vpc.vpc[0].crn : data.ibm_is_vpc.vpc[0].crn
```
ocofaigh, line 17: Does this `resource.ibm_is_vpc.vpc[0].crn` work?? I would usually expect just `ibm_is_vpc.vpc[0].crn`

ocofaigh (changes requested):
@Aayush-Abhyarthi Still alot of issues with this PR. I left some comments. Please reach out if you are unclear on anything

main.tf
Aashiq-J: There is no mention of `resource_group` being a parameter for data source
https://registry.terraform.io/providers/IBM-Cloud/ibm/latest/docs/data-sources/is_vpc

variables.tf
```diff
@@ -18,8 +31,9 @@ variable "region" {
 }
 
 variable "prefix" {
-  description = "The prefix that you would like to append to your resources. Explicitly set to null if you do not wish to use a prefix."
+  description = "The value that you would like to prefix to the name of the resources provisioned by this module. Explicitly set to null if you do not wish to use a prefix."
   type        = string
+  default     = null
```
ocofaigh, line 36: thoughts on maybe adding a default prefix value here instead of `null` to be consistent with `name` variable? Or should we default `name` to null? cc @toddgiguere

tests/existing-resources/README.md
```diff
@@ -0,0 +1,3 @@
+# Existing Resources For use in tests
+
+<!-- Add an example where existing resources can be passed as variables for local testing (not for use with tests) -->
```
ocofaigh, line 3: `not for use with tests` is confusing - we are using it for tests, but as a prereq deployment to test the existing resources example

ocofaigh (commented):
This is looking much better. I left some final minor comments. @toddgiguere would also like to get a second pari of eyes on it please

tests/existing-resources/README.md
```diff
@@ -0,0 +1,3 @@
+# Existing Resources For use in tests
```
SirSpidey, line 1:
```suggestion
# Existing resources to use in tests
```

SirSpidey: Just a capitalization comment. See Conall's comments about what the title should be.

variables.tf
Aashiq-J: we will need some more code modification to support adding default value for `prefix`, while adding a default value for `name` works fine.
Screenshot 2023-11-15 at 6 17 27 PM

ocofaigh: I was suggesting we either set `name` and `prefix` to `null`, or give both default string values

ocofaigh (commented):
Just need to resolve the comments around what default value to give `name` and `prefix`

variables.tf
Aashiq-J: If we are to set `name` and `prefix` to `null`, we will need to update the naming logic everywhere for the existing-vpc example to work.

ocofaigh:
@toddgiguere would you mind giving this a final review please?

toddgiguere (approved):

ocofaigh (approved):

Issue #678, opened 2023-12-01
Incorrect names generated if prefix is null


vburckhardt:
Some checks for null are missing through the code.

Issue #677, opened 2023-12-01
Give full control on naming of objects created by this module


vburckhardt:
### Description

vpc, subnets, etc
This will requires extra fields in the inputs to take the full name.

The existing behavior using prefix to auto-generate name should remain, this is an opt-in new capability.

PR #685, opened 2023-12-04
feat: enhance naming for vpc resources and fix prefix checks

rajatagarwal-ibm:
### Description

Addresses:
* https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/issues/678
* https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/issues/677

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [x] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


rajatagarwal-ibm:
Ran every test locally and also ran hub-spoke-manual-resolver apply (doesn't have test) - everything worked fine.

rajatagarwal-ibm:
Need to skip the upgrade test as there is a change of the vpc name in the `main.tf`

ocofaigh:
@rajatagarwal-ibm It seems you added several things since skipping the upgrade test so I'm not comfortable with it being skipped. Can you unskip it (use commit with `UNSKIP UPGRADE TEST`) and paste the results. We can then make a call on skipping it if no breaking changes are identified.

examples/existing_vpc/variables.tf
```diff
@@ -9,10 +9,17 @@ variable "region" {
...
+
 variable "vpc_id" {
   description = "The ID of the VPC where the VSI will be created."
   type        = string
 }
+
+variable "public_gateway_name" {
+  description = "The name of the public gateway"
+  type        = string
+}
```
ocofaigh, line 21: Instead of adding a new variable, can you just hard code the variable values in the main.tf of the existing_vpc example main.tf ?

examples/existing_vpc/main.tf
```diff
@@ -16,5 +16,6 @@ module "slz_vpc" {
   create_vpc          = false
   existing_vpc_id     = var.vpc_id
   create_subnets      = false
+  public_gateway_name = var.public_gateway_name
```
ocofaigh, line 19: hard code a name here instead of using variable value (perhaps use the var.prefix in the name)

variables.tf
```diff
@@ -3,11 +3,47 @@
 ##############################################################################
 
 variable "name" {
-  description = "The name to give the newly provisioned VPC. Only used if 'create_vpc' is true."
+  description = "The name to give the provisioned VPC. Only used if 'create_vpc' is true."
```
ocofaigh, line 6: hmm, name is also used for other resources, not just the VPC. I would update this to:

"The string to use for the naming of VPC resources."

I think you should also then add a new variable called `vpc_name` which can be used to explicitly name the VPC.

main.tf
```diff
@@ -34,13 +34,27 @@ locals {
...
+}
+
 ##############################################################################
 # Create new VPC
 ##############################################################################
 
 resource "ibm_is_vpc" "vpc" {
   count                       = var.create_vpc == true ? 1 : 0
-  name                        = var.prefix != null ? "${var.prefix}-${var.name}-vpc" : "${var.name}-vpc"
+  name                        = var.prefix != null ? "${var.prefix}-${var.name}-vpc" : var.name
```
ocofaigh, line 57: As per my other comment, var.name is used for other VPC resource naming. I would suggest you create a new variable called `var.vpc_name` (defaults to null) and regardless of the value of `var.prefix`, if the user passes a value for `var.vpc_name` the code should use that. Otherwise use "${var.prefix}-${var.name}-vpc" unless prefix is null, in which case use "${var.name}-vpc"

main.tf
```diff
@@ -100,7 +114,7 @@ resource "ibm_is_vpc" "vpc" {
 resource "ibm_is_vpc_dns_resolution_binding" "vpc_dns_resolution_binding_id" {
   count = (var.enable_hub == false && var.enable_hub_vpc_id) ? 1 : 0
 
-  name   = "${var.prefix}-dns-binding"
+  name   = var.prefix != null ? "${var.prefix}-dns-binding" : var.dns_binding_name
```
ocofaigh, line 117: Regardless of the value of `var.prefix`, if the user passes a value for `var.dns_binding_name` the code should use that. Otherwise use "${var.prefix}-${var.name}-dns-binding" unless prefix is null, in which case use "${var.name}-dns-binding"

main.tf
```diff
@@ -109,26 +123,17 @@ resource "ibm_is_vpc_dns_resolution_binding" "vpc_dns_resolution_binding_id" {
 
 resource "ibm_is_vpc_dns_resolution_binding" "vpc_dns_resolution_binding_crn" {
   count  = (var.enable_hub == false && var.enable_hub_vpc_crn) ? 1 : 0
-  name   = "${var.prefix}-dns-binding"
+  name   = var.prefix != null ? "${var.prefix}-dns-binding" : var.dns_binding_name
```
ocofaigh, line 126: Regardless of the value of `var.prefix`, if the user passes a value for `var.dns_binding_name` the code should use that. Otherwise use "${var.prefix}-${var.name}-dns-binding" unless prefix is null, in which case use "${var.name}-dns-binding"

main.tf
```diff
@@ -109,26 +123,17 @@ resource "ibm_is_vpc_dns_resolution_binding" "vpc_dns_resolution_binding_id" {
...
-
-locals {
-  vpc_id = var.create_vpc ? ibm_is_vpc.vpc[0].id : var.existing_vpc_id
-}
-
 # Configure custom resolver on the hub vpc
 resource "ibm_resource_instance" "dns_instance_hub" {
   count             = var.enable_hub && !var.skip_custom_resolver_hub_creation && !var.use_existing_dns_instance ? 1 : 0
-  name              = "${var.prefix}-dns-instance"
+  name              = var.prefix != null ? "${var.prefix}-dns-instance" : var.dns_instance_name
```
ocofaigh, line 136: Regardless of the value of `var.prefix`, if the user passes a value for `var.dns_instance_name` the code should use that. Otherwise use "${var.prefix}-${var.name}-dns-instance" unless prefix is null, in which case use "${var.name}-dns-instance"

main.tf
```diff
@@ -137,7 +142,7 @@ resource "ibm_resource_instance" "dns_instance_hub" {
 
 resource "ibm_dns_custom_resolver" "custom_resolver_hub" {
   count             = var.enable_hub && !var.skip_custom_resolver_hub_creation ? 1 : 0
-  name              = "${var.prefix}-custom-resolver"
+  name              = var.prefix != null ? "${var.prefix}-custom-resolver" : var.dns_custom_resolver_name
```
ocofaigh, line 145: Regardless of the value of `var.prefix`, if the user passes a value for `var.dns_custom_resolver_name` the code should use that. Otherwise use "${var.prefix}-${var.name}-custom-resolver" unless prefix is null, in which case use "${var.name}-custom-resolver"

main.tf
```diff
@@ -193,7 +198,7 @@ resource "time_sleep" "wait_for_authorization_policy" {
 
 resource "ibm_is_vpc_routing_table" "route_table" {
   for_each                      = module.dynamic_values.routing_table_map
-  name                          = var.prefix != null ? "${var.prefix}-${var.name}-route-${each.value.name}" : "${var.name}-route-${each.value.name}"
+  name                          = var.prefix != null ? "${var.prefix}-${var.name}-route-${each.value.name}" : "${var.routing_table_name}-${each.value.name}"
```
ocofaigh, line 201: I don't see any reason to add `var.routing_table_name` - you can achieve this naming with the current code using `var.name`. There can be multiple routing tables here, so we will always be appending `${each.value.name}` to the name here

main.tf
```diff
@@ -228,7 +233,7 @@ locals {
 
 resource "ibm_is_public_gateway" "gateway" {
   for_each       = local.gateway_object
-  name           = var.prefix != null ? "${var.prefix}-${var.name}-public-gateway-${each.key}" : "${var.name}-public-gateway-${each.key}"
+  name           = var.prefix != null ? "${var.prefix}-${var.name}-public-gateway-${each.key}" : "${var.public_gateway_name}-${each.key}"
```
ocofaigh, line 236: I don't see any reason to add `var.public_gateway_name` - you can achieve this naming with the current code using `var.name`. There can be multiple public gateways here, so we will always be appending `${each.value.name}` to the name here

variables.tf
```diff
@@ -3,11 +3,47 @@
...
+  default     = null
+}
+
+variable "dns_custom_resolver_name" {
+  description = "The name to give the provisioned DNS custom resolver instance. Only used if 'enable_hub' is true and 'skip_custom_resolver_hub_creation'is false."
+  type        = string
+  default     = null
+}
+
+variable "routing_table_name" {
```
ocofaigh, line 29: as per above comment, I propose to delete this. There can be multiple routing tables so var.name (and var.prefix is not null) can be used for the naming

variables.tf
```diff
@@ -3,11 +3,47 @@
...
+  description = "The name to give the provisioned Routing table."
+  type        = string
+  default     = null
+}
+
+variable "public_gateway_name" {
```
ocofaigh, line 35: as per above comment, I propose to delete this. There can be multiple public gateways so var.name (and var.prefix is not null) can be used for the naming

variables.tf
```diff
@@ -3,11 +3,47 @@
...
+  description = "The name to give the provisioned VPC Public Gateway."
+  type        = string
+  default     = null
+}
+
+variable "vpc_flow_logs_name" {
+  description = "The name to give the provisioned VPC flow logs."
```
ocofaigh, line 42: After you update the logic in my other comment, The description here should be updated to: 

"The name to give the provisioned VPC flow logs. If no value is used, a name will be generated using the `name` and `prefix` (if using) input variables."

ocofaigh (changes requested):
@rajatagarwal-ibm See my comments please

tests/pr_test.go
```diff
@@ -84,7 +84,7 @@ func TestRunNoPrefixExample(t *testing.T) {
 
 	// check if name is the same
 	outputs := terraform.OutputAll(options.Testing, options.TerraformOptions)
-	assert.Equal(t, testName+"-vpc", outputs["vpc_name"], "VPC name was altered, and should not have been.")
```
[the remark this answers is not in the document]
rajatagarwal-ibm: Yes, the assertion is working fine. I changed the assertion and it failed.

image

main.tf
rajatagarwal-ibm: Now changed it to the following. It will always check first if `dns_binding_name` is there or not, if no then check with prefix or name. 

```
name   = var.dns_binding_name != null ? var.dns_binding_name : (var.prefix != null ? "${var.prefix}-dns-binding" : "${var.name}-dns-binding")
```

main.tf
rajatagarwal-ibm: As the objective of the task is to give freedom to the user to name it anything they like, I have added these as variables. If they don't provide the value then it will pick up either prefix or name.

variables.tf
rajatagarwal-ibm: This module is being consumed in the SLZ module. Any changes which we make here with `name` will impact that too. I don't think so that's needed. Hence my proposal is I can rename the description to the following:

```
The string to use for the naming of VPC and its resources.
```

main.tf
ocofaigh: But how are you going to handle the naming of multiple routing tables with 1 string variable. You dont want them all named the same, so hence you will need to use `${each.value.name}` in the naming

main.tf
ocofaigh: same as https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/pull/685#discussion_r1426907203

main.tf
```diff
@@ -34,13 +34,27 @@ locals {
...
+}
+
 ##############################################################################
 # Create new VPC
 ##############################################################################
 
 resource "ibm_is_vpc" "vpc" {
   count                       = var.create_vpc == true ? 1 : 0
-  name                        = var.prefix != null ? "${var.prefix}-${var.name}-vpc" : "${var.name}-vpc"
+  name                        = var.name != null ? var.name : "${var.prefix}-vpc"
```
vburckhardt, line 57: This is going to break of name and prefix are both set to null. Perhaps add a check to ensure that one of them is set.

main.tf
```diff
@@ -100,16 +114,17 @@ resource "ibm_is_vpc" "vpc" {
...
     id = var.hub_vpc_id # Target VPC ID
   }
 }
 
 resource "ibm_is_vpc_dns_resolution_binding" "vpc_dns_resolution_binding_crn" {
-  count  = (var.enable_hub == false && var.enable_hub_vpc_crn) ? 1 : 0
-  name   = "${var.prefix}-dns-binding"
+  count = (var.enable_hub == false && var.enable_hub_vpc_crn) ? 1 : 0
+
+  name   = var.dns_binding_name != null ? var.dns_binding_name : (var.prefix != null ? "${var.prefix}-dns-binding" : "${var.name}-dns-binding")
```
vburckhardt, line 127: Using coalesce may make this easier to read. Also I'd suggest adding a comment above.

vburckhardt (commented):
Hi Rajat, some comments

examples/existing_vpc/main.tf
rajatagarwal-ibm: There is no var.prefix in this example. In the tests, I am passing like `fmt.Sprintf("%s-public-gateway", prefix),`

main.tf
```diff
@@ -194,8 +212,9 @@ resource "time_sleep" "wait_for_authorization_policy" {
 ##############################################################################
 
 resource "ibm_is_vpc_routing_table" "route_table" {
-  for_each                      = module.dynamic_values.routing_table_map
-  name                          = var.prefix != null ? "${var.prefix}-${var.name}-route-${each.value.name}" : "${var.name}-route-${each.value.name}"
+  for_each = module.dynamic_values.routing_table_map
+  # Use var.routing_table_name if not null, otherwise, use var.prefix or var.name.
+  name                          = var.routing_table_name != null ? "${var.routing_table_name}-${each.value.name}" : (var.prefix != null ? "${var.prefix}-route-${each.value.name}" : "${var.name}-route-${each.value.name}")
```
ocofaigh, line 217: If the consumer passed a value for `prefix` and for `name` (and no value for `routing_table_name`), then I think we will want the routing table name name to be called `${var.prefix}-${var.name}-route-${each.value.name}`. Currently logic does not take that into consideration.

main.tf
```diff
@@ -101,45 +122,42 @@ resource "ibm_is_vpc" "vpc" {
...
   plan              = var.dns_plan
 }
 
 resource "ibm_dns_custom_resolver" "custom_resolver_hub" {
-  count             = var.enable_hub && !var.skip_custom_resolver_hub_creation ? 1 : 0
-  name              = "${var.prefix}-custom-resolver"
+  count = var.enable_hub && !var.skip_custom_resolver_hub_creation ? 1 : 0
+
+  # Use var.dns_custom_resolver_name if not null, otherwise, use var.prefix or var.name.
+  name              = coalesce(var.dns_custom_resolver_name, "${var.prefix != null ? var.prefix : var.name}-custom-resolver")
```
ocofaigh, line 160: If the consumer passed a value for `prefix` and for `name` (and no value for `dns_custom_resolver_name`), then I think we will want the dns resolver name name to be called `${var.prefix}-${var.name}-custom-resolver`. Currently logic does not take that into consideration.

main.tf
```diff
@@ -101,45 +122,42 @@ resource "ibm_is_vpc" "vpc" {
...
-}
-
 # Configure custom resolver on the hub vpc
 resource "ibm_resource_instance" "dns_instance_hub" {
-  count             = var.enable_hub && !var.skip_custom_resolver_hub_creation && !var.use_existing_dns_instance ? 1 : 0
-  name              = "${var.prefix}-dns-instance"
+  count = var.enable_hub && !var.skip_custom_resolver_hub_creation && !var.use_existing_dns_instance ? 1 : 0
+
+  # Use var.dns_instance_name if not null, otherwise, use var.prefix or var.name.
+  name              = coalesce(var.dns_instance_name, "${var.prefix != null ? var.prefix : var.name}-dns-instance")
```
ocofaigh, line 149: If the consumer passed a value for `prefix` and for `name` (and no value for `dns_instance_name `), then I think we will want the dns instance name name to be called `${var.prefix}-${var.name}-dns-instance`. Currently logic does not take that into consideration.

main.tf
```diff
@@ -101,45 +122,42 @@ resource "ibm_is_vpc" "vpc" {
...
   }
 }
 
 resource "ibm_is_vpc_dns_resolution_binding" "vpc_dns_resolution_binding_crn" {
-  count  = (var.enable_hub == false && var.enable_hub_vpc_crn) ? 1 : 0
-  name   = "${var.prefix}-dns-binding"
+  count = (var.enable_hub == false && var.enable_hub_vpc_crn) ? 1 : 0
+
+  # Use var.dns_binding_name if not null, otherwise, use var.prefix or var.name.
+  name   = coalesce(var.dns_binding_name, "${var.prefix != null ? var.prefix : var.name}-dns-binding")
```
ocofaigh, line 137: If the consumer passed a value for `prefix` and for `name` (and no value for `dns_binding_name `), then I think we will want the dns binding name name to be called `${var.prefix}-${var.name}-dns-binding`. Currently logic does not take that into consideration.

main.tf
```diff
@@ -101,45 +122,42 @@ resource "ibm_is_vpc" "vpc" {
 resource "ibm_is_vpc_dns_resolution_binding" "vpc_dns_resolution_binding_id" {
   count = (var.enable_hub == false && var.enable_hub_vpc_id) ? 1 : 0
 
-  name   = "${var.prefix}-dns-binding"
+  # Use var.dns_binding_name if not null, otherwise, use var.prefix or var.name.
+  name   = coalesce(var.dns_binding_name, "${var.prefix != null ? var.prefix : var.name}-dns-binding")
```
ocofaigh, line 126: If the consumer passed a value for `prefix` and for `name` (and no value for `dns_binding_name `), then I think we will want the dns binding name name to be called `${var.prefix}-${var.name}-dns-binding`. Currently logic does not take that into consideration.

ocofaigh (changes requested):
@rajatagarwal-ibm I left some comments - maybe we need to jump on a call to discuss how we want this to look as its starting to get confusing now to support all the use cases

main.tf
```diff
@@ -34,17 +40,31 @@ locals {
...
+}
+
 ##############################################################################
 # Create new VPC
 ##############################################################################
 
 resource "ibm_is_vpc" "vpc" {
   count          = var.create_vpc == true ? 1 : 0
-  name           = var.prefix != null ? "${var.prefix}-${var.name}-vpc" : "${var.name}-vpc"
+  name           = "${var.prefix != null && var.name != null ? "${var.prefix}-${var.name}" : var.prefix != null ? var.prefix : var.name}-vpc"
```
ocofaigh, line 63: This logic still does not provide the ability to have full control over the VPC name - suggest to add optional `vpc_name` variable

ocofaigh, line 63: `var.prefix != null && var.name != null ? "${var.prefix}-${var.name}-vpc" : var.prefix != null ? "${var.prefix}-vpc" : var.name`

variables.tf
```diff
@@ -55,6 +43,58 @@ variable "access_tags" {
...
+##############################################################################
+
+variable "prefix" {
+  description = "The value that you would like to prefix to the name of the resources provisioned by this module. Explicitly set to null if you do not wish to use a prefix."
+  type        = string
+  default     = null
+}
+
+variable "name" {
+  description = "The string to use for the naming of VPC, when var.create_vpc is true. This string is also used as a prefix for the naming of VPC resources."
```
ocofaigh, line 57: If prefix value is not passed, then the value of this variable is exactly what the VPC will be called.

README.md
```diff
@@ -58,6 +58,10 @@ module vpc {
 }
 ```
 
+### Resource naming
+
+The module automatically generates names for the all provisioned VPC resources using the `var.prefix` input variable. You can selectively override this behavior by giving explicit names through the following input variables: `dns_binding_name`, `dns_instance_name`, `dns_custom_resolver_name`, `routing_table_name`, `public_gateway_name`, and `vpc_flow_logs_name`.
```
ocofaigh, line 63: var.name (for VPC name)

dynamic_values.tf
```diff
@@ -4,7 +4,7 @@
 
 module "dynamic_values" {
   source               = "./dynamic_values"
-  prefix               = var.prefix != null ? "${var.prefix}-${var.name}" : var.name
+  prefix               = var.prefix != null ? "${var.prefix}-vpc" : var.name
```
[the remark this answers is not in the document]
ocofaigh: `var.prefix != null && var.name != null ? "${var.prefix}-${var.name}" : var.prefix != null ? "${var.prefix}" : var.name`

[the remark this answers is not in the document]
ocofaigh: ^^ try this and paste the terraform plan that is generated

main.tf
```diff
@@ -34,17 +40,31 @@ locals {
...
   resource_group = var.resource_group_id
   classic_access = var.classic_access
   # address prefix is set to auto only if no address prefixes NOR any subnet is passed as input
-  address_prefix_management   = (length([for prefix in values(coalesce(var.address_prefixes, {})) : prefix if prefix != null]) != 0) || (length([for subnet in values(coalesce(var.subnets, {})) : subnet if subnet != null]) != 0) ? "manual" : null
+  address_prefix_management   = length([for prefix in values(coalesce(var.address_prefixes, {})) : prefix if prefix != null]) != 0 ? "manual" : null
```
ocofaigh, line 67: why is this change needed?

variables.tf
```diff
@@ -55,6 +43,58 @@ variable "access_tags" {
...
+  type        = string
+  default     = null
+}
+
+variable "dns_binding_name" {
+  description = "The name to give the provisioned VPC DNS resolution binding. If not set, the module generates a name based on the var.prefix variable."
```
ocofaigh, line 63: var.prefix and var.name variables

ocofaigh: repeat for others

variables.tf
```diff
@@ -55,6 +43,58 @@ variable "access_tags" {
...
+}
+
+variable "dns_custom_resolver_name" {
+  description = "The name to give the provisioned DNS custom resolver instance. If not set, the module generates a name based on the var.prefix variable."
+  type        = string
+  default     = null
+}
+
+variable "routing_table_name" {
+  description = "The name to give the provisioned routing table. If not set, the module generates a name based on the var.prefix variable."
```
ocofaigh, line 81: routing tables

variables.tf
```diff
@@ -55,6 +43,58 @@ variable "access_tags" {
...
+  type        = string
+  default     = null
+}
+
+variable "public_gateway_name" {
+  description = "The name to give the provisioned VPC public gateway. If not set, the module generates a name based on the var.prefix variable."
```
ocofaigh, line 87: public gateways

rajatagarwal-ibm:
Created https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/pull/697 since we can no longer bypass the upgrade test in that version.

This can be closed

Issue #757, opened 2024-03-29
ibm_is_vpc_dns_resolution_binding between VPCs now requires auth policy


toddgiguere:
The latest test of example `hub-spoke-delegated-resolver` is failing to set up the DNS resolver in the spoke VPC with the following error:

> Error: CreateVPCDnsResolutionBindingWithContext failed the provided token is not authorized to connect the specified dns-resolution-binding (ID:r006-96473c8b-81a7-4738-b588-e91601e70bb4) in this account

In the IBM Cloud Web UI for VPC, when setting up a DNS resolver manually, the following text is displayed:
> Important: The DNS hub VPC administrator must [create a service-to-service authentication policy](https://cloud.ibm.com/docs/vpc?topic=vpc-hub-spoke-s2s-auth&interface=ui) that grants this DNS-shared VPC DNSBindingConnector permission on the DNS hub VPC.

This might be a new requirement, as this example (and test case) have been running successfully up until March 19 2024, and only the most recent test run of the example (March 26) is reporting this error with no other changes to the module in that time.

PR #775, opened 2024-04-30
feat: add authorization policy for spoke -> hub DNS access

MatthewLemmond:
### Description
Create an authorization policy from the spoke vpc to hub vpc to authorize the spoke to receive the DNS information from the hub, more details on the issue in https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/issues/757

resolves https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/issues/757

### Release required?

- [ ] No release
- [x] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


examples/existing_vpc/variables.tf
```diff
@@ -32,5 +32,5 @@ variable "existing_resource_group_name" {
 variable "name" {
   description = "The string is used as a prefix for the naming of VPC resources."
   type        = string
-  default     = null
+  default     = "existing-vpc"
```
MatthewLemmond, line 35: As a note, pre-commit was failing on this line when it was null with the following:

```

TFLint in examples/existing_vpc/:
Failed to prepare rule checking; failed to eval an expression in ../../dynamic_values.tf:7; ../../dynamic_values.tf:7,64-72: Invalid template interpolation value; The expression result is null. Cannot include a null value in a string template.:

Error: Invalid template interpolation value

  on ../../dynamic_values.tf line 7, in module "dynamic_values":
   7:   prefix               = var.prefix != null ? "${var.prefix}-${var.name}" : var.name

with var.name set to null.

The expression result is null. Cannot include a null value in a string template.
```
I added a default string here to get past the pre-commit error, but this may need to have no default/require a value be passed in at all times

variables.tf
```diff
@@ -542,6 +542,18 @@ variable "enable_hub" {
   default     = false
 }
 
+variable "skip_spoke_auth_policy" {
+  description = "Set to true to skip the creation of an authorization policy between the DNS resolution spoke and hub, only enable this if a policy already exists between these two VPCs. See https://cloud.ibm.com/docs/vpc?topic=vpc-vpe-dns-sharing-s2s-auth&interface=ui for more details."
+  type        = bool
+  default     = false
+}
+
+variable "hub_account_id" {
```
rajatagarwal-ibm, line 551: Suggest you add a validation to ensure `hub_account_id` is passed when `skip_spoke_auth_policy` is false.

main.tf
```diff
@@ -116,9 +116,48 @@ resource "ibm_is_vpc" "vpc" {
...
 ##############################################################################
 
+# fetch this account ID
+data "ibm_iam_account_settings" "iam_account_settings" {}
+
+# spoke -> hub auth policy based on https://cloud.ibm.com/docs/vpc?topic=vpc-vpe-dns-sharing-s2s-auth&interface=terraform
+resource "ibm_iam_authorization_policy" "vpc_dns_resolution_auth_policy" {
+  count = (var.enable_hub == false && var.skip_spoke_auth_policy == false && (var.enable_hub_vpc_id || var.enable_hub_vpc_crn)) ? 1 : 0
+  roles = ["DNS Binding Connector"]
+  subject_attributes {
```
rajatagarwal-ibm, line 126: Perhaps add comment to explain "subject" is Spoke VPC and "resource" is Hub VPC. It will make easier for understanding for the maintainer.

examples/existing_vpc/variables.tf
rajatagarwal-ibm: I'd suggest to make it mandatory

rajatagarwal-ibm:
Also, you could update readme in examples and at the root module about supporting different accounts for hub and spoke vpcs.

rajatagarwal-ibm (changes requested):

MatthewLemmond:
The hub and spoke delegated example (weekly test) is failing with a different than the one seen that prompted this PR, error is as follows: 
```
Error updating the custom resolver to disable before deleting
Not allowed to disable custom resolver on hub VPC which has DNS resolution bindings.
```
Not sure if this is an issue with the terraform or a problem on the provider side but needs more investigation

rajatagarwal-ibm (approved):

rajatagarwal-ibm:
> The hub and spoke delegated example (weekly test) is failing with a different error than the one seen that prompted this PR, error is as follows:
> 
> ```
> Error updating the custom resolver to disable before deleting
> Not allowed to disable custom resolver on hub VPC which has DNS resolution bindings.
> ```
> 
> Not sure if this is an issue with the terraform or a problem on the provider side but needs more investigation

Would you like a create an issue please? I think there might be some update to the provider and we may need to investigate the deletion flow.

MatthewLemmond:
https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/issues/776 created to track the aforementioned issue with the destroy

PR #915, opened 2025-02-14
feat: initial commit for DA

maheshwarishikha:
### Description

VPC DA - fully configurable variant

Issue: https://github.ibm.com/GoldenEye/issues/issues/12217

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [x] Major release (`X.x.x`)

##### Release notes content

- Fully configurable variant
- Adds vpn_gateways feature in base module

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


solutions/simple/main.tf
```diff
@@ -0,0 +1,128 @@
...
+  existing_resource_group_name = var.use_existing_resource_group == true ? var.resource_group_name : null
+}
+
+#############################################################################
+# COS Bucket for VPC flow logs
+#############################################################################
+
+# parse COS details from the existing COS instance CRN
+module "existing_cos_crn_parser" {
+  count   = var.existing_cos_instance_crn != null ? 1 : 0
```
ocofaigh, line 22: using a CRN around a count can lead into issues where count is not known before apply. I suspect however we will never hit this in a DA unless someone tries to consume it as a module where they also create the COS instance in the same state file

solutions/simple/main.tf
```diff
@@ -0,0 +1,128 @@
...
+    management_endpoint_type      = var.management_endpoint_type_for_bucket
+    storage_class                 = var.cos_bucket_class
+    resource_instance_id          = var.existing_cos_instance_crn
+    region_location               = var.region
+    force_delete                  = true
+  }]
+}
+
+module "cos_buckets" {
+  count          = var.enable_vpc_flow_logs ? 1 : 0
```
ocofaigh, line 47: are we supporting user bringing there own bucket?

solutions/simple/variables.tf
```diff
@@ -0,0 +1,400 @@
...
+variable "use_existing_resource_group" {
+  type        = bool
+  description = "Whether to use an existing resource group."
+  default     = true
+}
+
+variable "resource_group_name" {
+  type        = string
+  description = "The name of a new or an existing resource group to provision the watsonx.ai resources. If a prefix input variable is specified, the prefix is added to the name in the `<prefix>-<name>` format."
+  default     = "Default"
```
ocofaigh, line 31: If we want to be consistent with other DAs, this should have no default value

solutions/simple/variables.tf
```diff
@@ -0,0 +1,400 @@
...
+
+variable "prefix" {
+  type        = string
+  description = "Prefix to add to all the resources created by this solution. To not use any prefix value, you can set this value to `null` or an empty string."
+  default     = "dev"
+}
+
+variable "vpc_name" {
+  default     = "simple"
+  description = "Name of the VPC."
```
ocofaigh, line 42: description needs to mention the prefix variable

solutions/simple/variables.tf
```diff
@@ -0,0 +1,400 @@
...
+
+variable "resource_tags" {
+  description = "Optional list of tags for the resources created by this solution."
+  type        = list(string)
+  default     = []
+}
+
+variable "access_tags" {
+  type        = list(string)
+  description = "A list of access tags to apply to the VPC resources created by the module. For more information, see https://cloud.ibm.com/docs/account?topic=account-access-tags-tutorial."
```
ocofaigh, line 60: Its not a module, so don't use the word "module" in descrption

solutions/simple/variables.tf
```diff
@@ -0,0 +1,400 @@
...
+    error_message = "Tags must match the regular expression \"[\\w\\-_\\.]+:[\\w\\-_\\.]+\". For more information, see https://cloud.ibm.com/docs/account?topic=account-tag&interface=ui#limits."
+  }
+}
+
+##############################################################################
+# Subnets
+##############################################################################
+
+variable "subnets" {
+  description = "List of subnets for the vpc. For each item in each array, a subnet will be created. Items can be either CIDR blocks or total ipv4 addressess. Public gateways will be enabled only in zones where a gateway has been created"
```
ocofaigh, line 76: We should add a supporting markdown doc

solutions/simple/variables.tf
```diff
@@ -0,0 +1,400 @@
...
+    ]
+  }
+}
+
+##############################################################################
+# Network ACLs
+##############################################################################
+
+variable "network_acls" {
+  description = "The list of ACLs to create. Provide at least one rule for each ACL."
```
ocofaigh, line 119: We should add a supporting markdown doc

solutions/simple/variables.tf
```diff
@@ -0,0 +1,400 @@
...
+# VPC Flow Logs Variables
+##############################################################################
+
+variable "enable_vpc_flow_logs" {
+  description = "Flag to enable vpc flow logs. If true, flow log collector will be created"
+  type        = bool
+  default     = false
+}
+
+variable "create_authorization_policy_vpc_to_cos" {
```
ocofaigh, line 298: keep consistent with other DAs - use the `skip_` naming convention

solutions/simple/variables.tf
```diff
@@ -0,0 +1,400 @@
...
+  description = "Create authorisation policy for VPC to access COS. Set as false if authorization policy exists already"
+  type        = bool
+  default     = true
+}
+
+variable "existing_cos_instance_crn" {
+  description = "GUID of the COS instance to create Flow log collector"
```
ocofaigh, line 305: GUID -> CRN

solutions/simple/version.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  required_version = ">= 1.3.0"
```
ocofaigh, line 2: you should bump this to >= 1.9.0 and use cross variable validation. There are many combinations / permutations so we badly need to use that kind of validation in the variables.

ocofaigh (commented):
I know it is still under development, but I left a few comments to make sure we address them. We should talk about the variation name. `simple` is probably not what we will go with

solutions/simple/main.tf
maheshwarishikha: No, in simple variation...I planned to skip supporting existing storage bucket. So user will bring the COS instance and a new bucket will always be created.

We can discuss...currently it is implemented like this.

solutions/simple/variables.tf
maheshwarishikha: yes, that was by mistake. Modified.

solutions/simple/variables.tf
maheshwarishikha: updated.

solutions/simple/variables.tf
maheshwarishikha: updated.

maheshwarishikha:
> I know it is still under development, but I left a few comments to make sure we address them. We should talk about the variation name. `simple` is probably not what we will go with

I created this draft PR with the intent of early review. Thanks for this. Yes, name will be changed once we decide.

solutions/simple/variables.tf
maheshwarishikha: will add, once code is completed.

maheshwarishikha:
Currently,
- this simple variation does not support `existing_cos_kms_key_crn` scenario.
- it does not support the existing COS buckets. 
More secured version can have both of these. Let me know your thoughts.

solutions/simple/version.tf
maheshwarishikha: done

solutions/simple/variables.tf
maheshwarishikha: done

solutions/simple/version.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  required_version = ">= 1.9.0"
+  required_providers {
+    # Use "greater than or equal to" range in modules
+    ibm = {
+      source  = "IBM-Cloud/ibm"
+      version = ">= 1.59.0, < 2.0.0"
```
ocofaigh, line 7: Lock into exact version for a DA

solutions/simple/version.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  required_version = ">= 1.9.0"
+  required_providers {
+    # Use "greater than or equal to" range in modules
```
ocofaigh, line 4: `# Lock DA into an exact provider version - renovate automation will keep it updated`

solutions/simple/variables.tf
```diff
@@ -0,0 +1,399 @@
...
+}
+
+variable "default_security_group_name" {
+  description = "OPTIONAL - Name of the Default Security Group. If null, a name will be automatically generated"
+  type        = string
+  default     = null
+}
+
+variable "default_routing_table_name" {
+  description = "OPTIONAL - Name of the Default Routing Table. If null, a name will be automatically generated"
```
ocofaigh, line 396: The guidance from docs team is not to add `OPTIONAL` in the variable description. In projects it will show in the optional tab, so no need to duplicate that

ocofaigh (changes requested):
We really need to make use of cross variable validation in the variables.tf file. For example, if person sets `kms_encryption_enabled_bucket` or `enable_vpc_flow_logs` then other inputs become required. 

Also question - should we support creating VPEs? Or are we planning to have a standalone VPE DA? (cc @vburckhardt )

ocofaigh:
@maheshwarishikha FYI these are the variation names we wil go with:

1. label: `Fully configurable`, name: `fully-configurable`
2. label: Security-enforced, name: `security-enforced`

solutions/simple/variables.tf
maheshwarishikha: done

solutions/simple/version.tf
maheshwarishikha: added

solutions/simple/version.tf
maheshwarishikha: addressed

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,473 @@
...
+}
+
+variable "prefix" {
+  type        = string
+  description = "Prefix to add to all the resources created by this solution. To not use any prefix value, you can set this value to `null` or an empty string."
+  default     = "dev"
+}
+
+variable "vpc_name" {
+  default     = "simple"
```
ocofaigh, line 40: we probably need a better default than this? Perhaps it should be "vpc"? So you would end up with "dev-vpc" when using prefix?

maheshwarishikha: updated

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+}
+
+variable "cos_bucket_name" {
+  description = "Name of the Cloud Object Storage bucket to be created collect VPC flow logs."
+  type        = string
+  default     = "cos-bucket"
+}
+
+variable "kms_encryption_enabled_bucket" {
+  description = "Set to true if Cloud Object Storage bucket needs to be KMS encryption enabled."
```
ocofaigh, line 373: Set to true to encrypt the Cloud Object Storage Flow Logs bucket with a KMS key. If set to true, a value must be passed for `existing_kms_key_crn` (to use that key) or `existing_kms_instance_crn` (to create a new key). Value cannot be set to true if `enable_vpc_flow_logs` is set to false.

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+    condition     = contains(["standard", "vault", "cold", "smart", "onerate_active"], var.cos_bucket_class)
+    error_message = "Specify one of the following values for the `cos_bucket_class`: `standard`, `vault`, `cold`, `smart`, or `onerate_active`."
+  }
+}
+
+###############################################################################################################
+# KMS
+###############################################################################################################
+
+variable "existing_kms_key_crn" {
```
ocofaigh, line 418: I actually suggest you name this `existing_flow_logs_bucket_kms_key_crn` so it is clear what the key is used for. Maybe some day there be something else in this DA that will need its own KMS key, so better to future proof it

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+variable "ibmcloud_api_key" {
+  type        = string
+  description = "The IBM Cloud API key to deploy resources."
+  sensitive   = true
+}
+
+variable "provider_visibility" {
+  description = "Set the visibility value for the IBM terraform provider. Supported values are `public`, `private`, `public-and-private`. [Learn more](https://registry.terraform.io/providers/IBM-Cloud/ibm/latest/docs/guides/custom-service-endpoints)."
+  type        = string
+  default     = "public"
```
ocofaigh, line 14: set this to private

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+
+variable "existing_resource_group_name" {
+  type        = string
+  description = "The name of an existing resource group to provision the resources."
+}
+
+variable "prefix" {
+  type        = string
+  nullable    = true
+  description = "Prefix to add to all the resources created by this solution. To not use any prefix value, you can set this value to `null` or an empty string."
```
ocofaigh, line 30: since this DA support creating a COS bucket, you should add extra validation on the value. See https://github.com/terraform-ibm-modules/terraform-ibm-scc-da/blob/49ba987b2679e8e8bae0af451922ec525252fe92/solutions/instances/variables.tf#L33-L40

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+  type        = string
+}
+
+variable "region" {
+  default     = "us-south"
+  description = "Region to deploy the VPC."
+  type        = string
+}
+
+variable "resource_tags" {
```
ocofaigh, line 45: different vpc resource support tags (e.g. a VPC can have different tags to subnets). I think this needs to be broken down. VPC tags variable should be explicitly named `vpc_resource_tags`

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+  }
+}
+
+##############################################################################
+# Subnets
+##############################################################################
+
+variable "subnets" {
+  description = "List of subnets for the vpc. For each item in each array, a subnet will be created. Items can be either CIDR blocks or total ipv4 addressess. Public gateways will be enabled only in zones where a gateway has been created"
+  type = object({
```
ocofaigh, line 70: I suggest to add a `subnet_resource_tags` attribute so subnets can individually tagged

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+      ])
+    )) == 0
+  }
+
+}
+
+variable "clean_default_sg_acl" {
+  description = "Remove all rules from the default VPC security group and VPC ACL (less permissive)"
+  type        = bool
+  default     = false
```
ocofaigh, line 288: hmm, should this default to secure by design (aka true) ?

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+}
+
+variable "skip_vpc_cos_authorization_policy" {
+  description = "To skip creating an IAM authorization policy that allows the VPC to access the Cloud Object Storage, set this variable to `true`."
+  type        = bool
+  default     = true
+}
+
+variable "existing_cos_instance_crn" {
+  description = "CRN of the existing COS instance. It is required to create the bucket used for flow logs."
```
ocofaigh, line 356: You should mention it is only required if `enable_vpc_flow_logs` is set to true in the description

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+  type        = string
+  default     = null
+
+  validation {
+    condition     = var.enable_vpc_flow_logs ? (var.existing_cos_instance_crn != null ? true : false) : true
+    error_message = "'existing_cos_instance_crn' is required if 'enable_vpc_flow_logs' is set to true."
+  }
+}
+
+variable "cos_bucket_name" {
```
ocofaigh, line 366: perhaps call it `flow_logs_new_bucket_name` so its clear that its a new bucket that will be created for flow logs

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+  description = "Name of the Cloud Object Storage bucket to be created collect VPC flow logs."
+  type        = string
+  default     = "cos-bucket"
```
ocofaigh, line 369: maybe rename this to flow-logs-bucket

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+variable "skip_cos_kms_auth_policy" {
+  type        = bool
+  description = "To skip creating an IAM authorization policy that allows Cloud Object Storage(COS) to access KMS key."
+  default     = false
+}
+
+variable "management_endpoint_type_for_bucket" {
+  description = "The type of endpoint for the IBM Terraform provider to use to manage Cloud Object Storage buckets (`public`, `private`, or `direct`). If you are using a private endpoint, make sure that you enable virtual routing and forwarding (VRF) in your account, and that the Terraform runtime can access the IBM Cloud Private network."
+  type        = string
+  default     = "public"
```
ocofaigh, line 397: default this to direct

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+variable "existing_kms_instance_crn" {
+  type        = string
+  default     = null
+  description = "The CRN of the existing key management service (KMS) that is used to create keys for encrypting the Cloud Object Storage bucket."
+}
+
+variable "kms_endpoint_type" {
+  type        = string
+  description = "The type of endpoint to use for communicating with the Key Protect instance. Possible values: `public`, `private`. Applies only if `existing_cos_kms_key_crn` is not specified."
+  default     = "public"
```
ocofaigh, line 433: private

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+  validation {
+    condition     = can(regex("public|private", var.kms_endpoint_type))
+    error_message = "Valid values for the `kms_endpoint_type_value` are `public` or `private`."
+  }
+}
+
+variable "kms_key_ring_name" {
```
ocofaigh, line 440: should we be more explicit? `flow_logs_bucket_kms_key_ring_name` ?

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+  type        = string
+  default     = "cos-key-ring"
```
ocofaigh, line 442: flow-logs-bucket-key-ring

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+  description = "The name of the key ring to create for the Cloud Object Storage bucket key. If an existing key is used, this variable is not required. If the prefix input variable is passed, the name of the key ring is prefixed to the value in the `<prefix>-value` format."
+}
+
+variable "kms_key_name" {
```
ocofaigh, line 446: should we be more explicit? flow_logs_bucket_kms_key_name ?

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,492 @@
...
+  type        = string
+  default     = "cos-key"
```
ocofaigh, line 448: flow-logs-bucket-key

ocofaigh:
Should we be supporting existing VPC @vburckhardt ?

solutions/fully-configurable/variables.tf
maheshwarishikha: updated

solutions/fully-configurable/variables.tf
maheshwarishikha: updated

solutions/fully-configurable/variables.tf
maheshwarishikha: added

solutions/fully-configurable/variables.tf
maheshwarishikha: I think, we can make it true for security-enforced variant. what say?

solutions/fully-configurable/variables.tf
maheshwarishikha: should we make this change only in DA or in module?

solutions/fully-configurable/variables.tf
maheshwarishikha: added

solutions/fully-configurable/variables.tf
maheshwarishikha: Modified as `flow_logs_cos_bucket_name` , description says bucket will be created.

solutions/fully-configurable/variables.tf
maheshwarishikha: done

solutions/fully-configurable/variables.tf
maheshwarishikha: done

solutions/fully-configurable/variables.tf
maheshwarishikha: The customers create separate keys for different purposes. Is the creation of different key-rings recommended?

solutions/fully-configurable/main.tf
```diff
@@ -0,0 +1,136 @@
...
+    bucket_name                   = local.bucket_name
+    kms_encryption_enabled        = var.kms_encryption_enabled_bucket
+    kms_guid                      = var.kms_encryption_enabled_bucket ? module.existing_kms_instance_crn_parser[0].service_instance : null
+    kms_key_crn                   = var.kms_encryption_enabled_bucket ? var.existing_kms_instance_crn : null
+    skip_iam_authorization_policy = var.skip_cos_kms_auth_policy
+    management_endpoint_type      = var.management_endpoint_type_for_bucket
+    storage_class                 = var.cos_bucket_class
+    resource_instance_id          = var.existing_cos_instance_crn
+    region_location               = var.region
+    force_delete                  = true
```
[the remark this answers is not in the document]
maheshwarishikha: exposed via the variable

solutions/fully-configurable/variables.tf
maheshwarishikha: done

solutions/fully-configurable/variables.tf
maheshwarishikha: done

solutions/fully-configurable/variables.tf
ocofaigh: I think the best practise for key rings is to split them by functional area. So maybe this is fine to leave as `kms_key_ring_name` and perhaps we default it to something like `vpc-key-ring` ?

solutions/fully-configurable/variables.tf
ocofaigh: as per above [comment](https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/pull/915#discussion_r1991378922), maybe go with `vpc-key-ring` ?

solutions/fully-configurable/variables.tf
ocofaigh: I think if set to false, it will fail one of the SCC rules as the rules are wide open. But if we set it to false, anything that is using the default ACL or SG will fail as it will have no rules, so no communication can occur unless user sets up an alternative ACL / SG. @toddgiguere any thoughts?

toddgiguere: I think having the default as "true" is only an issue when the subnets have the default ACL attached. I think that might be impossible in this case? If so then the default ACLs are orphans anyway (not attached to any subnets) so if they are empty (even accidentally) it won't matter? I'm thinking from ACL standpoint it might be safe to default to `true`.

Security Group is a little more tricky. By default it is (probably) not attached to anything in this module, but it is still out there and available, but I'm not sure how many would use the default sg, rather than create a new one.

To be honest I don't really see this as a "security enforcement", because these are not used or attached to anything so setting this to "true" isn't really affecting any traffic. We really added this because SCC is not smart enough to figure out that the ACLs are not attached to anything, and flags them as "too open" anyway.

maheshwarishikha:
**default inbound rules:**

image

**default outbound rules:**

image

solutions/fully-configurable/variables.tf
maheshwarishikha: Got the point. thanks. 
Made it true.

maheshwarishikha:
@ocofaigh @toddgiguere Please review this PR.

Few points/questions :
1. Have added all the required features except separate resource tags functionality for vpc, subnets, vpn gateways etc. Working on this. One question - as we plan to change variable names in module, will it be a breaking change for landing zone?

2. I have added VPE support but it is being attached with default security groups. Current code does not support creating more security groups as discussed before.

3. After adding VPE, `tf plan` shows `update in-place`. And `tf apply` gave error as:
```
Error: Error Deleting Security Group Target for this endpoint gateway : The specified endpoint gateway is not attached to any other security groups.
```
Is it a known issue? Seems to me a provider issue.

4. Docs for complex objects will be added soon.

ocofaigh:
> 1. Have added all the required features except separate resource tags functionality for vpc, subnets, vpn gateways etc. Working on this. One question - as we plan to change variable names in module, will it be a breaking change for landing zone?

^ Renaming variables is not a breaking change. Its a minor version change with release notes that clearly state the renamed vars. See this [doc](https://terraform-ibm-modules.github.io/documentation/#/versioning?id=major-version-changes) for breaking change criteria

> 2. I have added VPE support but it is being attached with default security groups. Current code does not support creating more security groups as discussed before.

^ Remind me why we are not supporting creation of more security groups? I would of thought this needs to be a feature of a VPC DA?

> 3. After adding VPE, tf plan shows update in-place. And tf apply gave error

^ How do we handle that provider bug in any of our other modules / examples that create VPE? Can we follow up with Ujjwal Kumar ?

> 4. Docs for complex objects will be added soon.

^ ok

> 5. Currently I have not added the following for cos buckets. Let me know if need to add this too.

^ This COS bucket will be used for VPC flow logs. So by default it should create an audit type bucket (retention enabled by default etc)

maheshwarishikha:
> This COS bucket will be used for VPC flow logs. So by default it should create an audit type bucket (retention enabled by default etc)

Added archive, expire, retention rules and object versioning.

> Regarding renaming of variables - tags related.

We have tags for vpc, subnet, flow logs, vpe gateway, vpn gateway, public gateway - should we really plan to add resource_tags for each?? Currently we use `tags` variable for each.

> Remind me why we are not supporting creation of more security groups? I would of thought this needs to be a feature of a VPC DA?

Previously we thought that it is an advanced feature, but that discussion was for quickstart. Added it now.

> How do we handle that provider bug in any of our other modules / examples that create VPE? 

Having default value as `null` instead of `[]` is helping. Made the same change in code.

maheshwarishikha:
FullyConfigurable upgrade test is not there in main branch, hence skipping upgrade test.

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,679 @@
...
+  type        = string
+  default     = "standard"
+  description = "The storage class of the newly provisioned Cloud Object Storage bucket. Specify one of the following values for the storage class: `standard`, `vault`, `cold`, `smart` (default), or `onerate_active`."
+  validation {
+    condition     = contains(["standard", "vault", "cold", "smart", "onerate_active"], var.cos_bucket_class)
+    error_message = "Specify one of the following values for the `cos_bucket_class`: `standard`, `vault`, `cold`, `smart`, or `onerate_active`."
+  }
+}
+
+variable "force_delete" {
```
ocofaigh, line 486: Don't expose this - it will prevent destroy. See https://ibm-cloudplatform.slack.com/archives/CG06FQ31P/p1741861817152299

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,679 @@
...
+variable "add_bucket_name_suffix" {
+  type        = bool
+  description = "Add a randomly generated suffix that is 4 characters in length, to the name of the newly provisioned Cloud Object Storage bucket. Do not use this suffix if you are passing the existing Cloud Object Storage bucket. To manage the name of the Cloud Object Storage bucket manually, use the `flow_logs_cos_bucket_name` variables."
+  default     = true
+}
+
+variable "flow_logs_cos_bucket_archive_days" {
+  description = "The number of days before the `archive_type` rule action takes effect for the flow logs cloud object storage bucket."
+  type        = number
+  default     = 20
```
ocofaigh, line 501: curious where this number came from?

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,679 @@
...
+}
+
+###############################################################################################################
+# KMS
+###############################################################################################################
+
+variable "existing_flow_logs_bucket_kms_key_crn" {
+  type        = string
+  default     = null
+  description = "The CRN of the existing root key of key management service (KMS) that is used to encrypt the flow logs Cloud Object Storage bucket."
```
ocofaigh, line 564: the description should mention to use `existing_kms_instance_crn` to create a new key

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,679 @@
...
+  )
+
+  default = []
+}
+
+##############################################################################
+# VPE Gateways
+##############################################################################
+
+variable "cloud_services" {
```
ocofaigh, line 641: suggest to prefix this with `vpe_gateway_`

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,679 @@
...
+  description = "The list of cloud services used to create endpoint gateways. If `vpe_name` is not specified in the list, VPE names are created in the format `<prefix>-<vpc_name>-<service_name>`."
+  type = set(object({
+    service_name                 = string
+    vpe_name                     = optional(string), # Full control on the VPE name. If not specified, the VPE name will be computed based on prefix, vpc name and service name.
+    allow_dns_resolution_binding = optional(bool, false)
+  }))
+  default = []
+}
+
+variable "cloud_service_by_crn" {
```
ocofaigh, line 651: suggest to prefix this with `vpe_gateway_`

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,679 @@
...
+
+  default = []
+}
+
+##############################################################################
+# VPE Gateways
+##############################################################################
+
+variable "cloud_services" {
+  description = "The list of cloud services used to create endpoint gateways. If `vpe_name` is not specified in the list, VPE names are created in the format `<prefix>-<vpc_name>-<service_name>`."
```
ocofaigh, line 642: The description needs to point to a helper markdown doc

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,679 @@
...
+  type = set(object({
+    service_name                 = string
+    vpe_name                     = optional(string), # Full control on the VPE name. If not specified, the VPE name will be computed based on prefix, vpc name and service name.
+    allow_dns_resolution_binding = optional(bool, false)
+  }))
+  default = []
+}
+
+variable "cloud_service_by_crn" {
+  description = "The list of cloud service CRNs used to create endpoint gateways. Use this list to identify services that are not supported by service name in the `cloud_services` variable. For a list of supported services, see [VPE-enabled services](https://cloud.ibm.com/docs/vpc?topic=vpc-vpe-supported-services). If `service_name` is not specified, the CRN is used to find the name. If `vpe_name` is not specified in the list, VPE names are created in the format `<prefix>-<vpc_name>-<service_name>`."
```
ocofaigh, line 652: The description needs to point to a helper markdown doc

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,679 @@
...
+      crn                          = string
+      vpe_name                     = optional(string) # Full control on the VPE name. If not specified, the VPE name will be computed based on prefix, vpc name and service name.
+      service_name                 = optional(string) # Name of the service used to compute the name of the VPE. If not specified, the service name will be obtained from the crn.
+      allow_dns_resolution_binding = optional(bool, true)
+    })
+  )
+  default = []
+}
+
+variable "vpe_service_endpoints" {
```
ocofaigh, line 664: suggest to prefix this with `vpe_gateway_`

ocofaigh (changes requested):
a few comments

solutions/fully-configurable/variables.tf
maheshwarishikha: I took this number from the default value defined in module - https://github.com/terraform-ibm-modules/terraform-ibm-cos/blob/main/modules/buckets/variables.tf#L36

solutions/fully-configurable/variables.tf
maheshwarishikha: updated

solutions/fully-configurable/variables.tf
maheshwarishikha: updated

solutions/fully-configurable/variables.tf
maheshwarishikha: updated

solutions/fully-configurable/variables.tf
maheshwarishikha: sure, removed.

maheshwarishikha:
> We have tags for vpc, subnet, flow logs, vpe gateway, vpn gateway, public gateway - should we really plan to add resource_tags for each?? Currently we use tags variable for each.

Discussed with Conall, the plan to do the following:
Use `var.tags` for all VPC related resources created by the DA. For extra subnet related tags, just add 1 more attribute in var.subnets to support adding different tags per subnet zone

variables.tf
```diff
@@ -718,3 +718,28 @@ variable "dns_records" {
     error_message = "Invalid MX record configuration. For 'MX' records, value for 'preference' must be provided."
   }
 }
+
+##############################################################################
+# VPN Gateways
+##############################################################################
+
+variable "enable_vpn_gateways" {
```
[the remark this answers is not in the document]
maheshwarishikha: thanks for the suggestion, updated code.

solutions/fully-configurable/variables.tf
maheshwarishikha: updated.

solutions/fully-configurable/variables.tf
maheshwarishikha: added.

solutions/fully-configurable/variables.tf
maheshwarishikha: added

solutions/simple/variables.tf
maheshwarishikha: added

rajatagarwal-ibm (changes requested):

ibm_catalog.json
```diff
@@ -0,0 +1,533 @@
...
+                  "version_input": "prefix",
+                  "reference_version": true
+                },
+                {
+                  "dependency_input": "region",
+                  "version_input": "region",
+                  "reference_version": true
+                },
+                {
+                  "dependency_input": "use_existing_resource_group",
```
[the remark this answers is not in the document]
Khuzaima05: @rajatagarwal-ibm use_existing_resource_group variable is part of kms add-on , here we are passing value to that variable.

solutions/fully-configurable/variables.tf
maheshwarishikha: As discussed, subnet_tags has been added explicitly.

solutions/fully-configurable/main.tf
```diff
@@ -0,0 +1,245 @@
...
+  count   = var.existing_cos_instance_crn != null ? 1 : 0
+  source  = "terraform-ibm-modules/common-utilities/ibm//modules/crn-parser"
+  version = "1.1.0"
+  crn     = var.existing_cos_instance_crn
+}
+
+locals {
+  cos_instance_guid                        = var.existing_cos_instance_crn != null ? module.existing_cos_crn_parser[0].service_instance : null
+  cos_account_id                           = var.existing_cos_instance_crn != null ? module.existing_cos_crn_parser[0].account_id : null
+  bucket_name                              = "${local.prefix}${var.flow_logs_cos_bucket_name}"
```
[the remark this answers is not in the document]
maheshwarishikha: This is correct. Its a new change `prefix` related. Now `local.prefix` has `-` already.

solutions/fully-configurable/main.tf
```diff
@@ -0,0 +1,245 @@
...
+
+  # configuration for the flow logs bucket
+  bucket_config = [{
+    access_tags                   = var.access_tags
+    bucket_name                   = local.bucket_name
+    add_bucket_name_suffix        = var.add_bucket_name_suffix
+    kms_encryption_enabled        = var.kms_encryption_enabled_bucket
+    kms_guid                      = local.kms_guid
+    kms_key_crn                   = local.cos_kms_key_crn
+    skip_iam_authorization_policy = local.create_cross_account_cos_kms_auth_policy || var.skip_cos_kms_iam_auth_policy
```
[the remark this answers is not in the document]
maheshwarishikha: `local.create_cross_account_cos_kms_auth_policy` will be true incase of cross account and if skip** is true then it means Buckets module wont create the policy. So I think it is fine.

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,694 @@
...
+  type        = bool
+  nullable    = false
+  description = "To skip creating an IAM authorization policy that allows Cloud Object Storage(COS) to access KMS key."
+  default     = false
+}
+
+variable "management_endpoint_type_for_bucket" {
+  description = "The type of endpoint for the IBM Terraform provider to use to manage Cloud Object Storage buckets (`public`, `private`, or `direct`). If you are using a private endpoint, make sure that you enable virtual routing and forwarding (VRF) in your account, and that the Terraform runtime can access the IBM Cloud Private network."
+  type        = string
+  default     = "direct"
```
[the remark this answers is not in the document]
maheshwarishikha: It is as per the suggestion provided by Conall. For DAs, we are keeping it direct everywhere.

ibm_catalog.json
```diff
@@ -0,0 +1,533 @@
...
+      ],
+      "flavors": [
+        {
+          "label": "Fully configurable",
+          "name": "fully-configurable",
+          "install_type": "fullstack",
+          "working_directory": "solutions/fully-configurable",
+          "architecture": {
+            "descriptions": "This architecture creates a VPC.",
+            "features": [
```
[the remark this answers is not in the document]
Khuzaima05: Updated the architecture description and acronyms. All add-on-related features are already covered at the top-level feature, so we are not including them in the flavor-specific features to avoid duplication.

.catalog-onboard-pipeline.yaml
```diff
@@ -0,0 +1,14 @@
...
+    kind: solution
+    catalog_id: f64499c8-eb50-4985-bf91-29f9e605a433
+    offering_id: ae881fc1-273f-472e-8437-99099a6a753b
+    include_git_submodules: true
+    variations:
+      - name: fully-configurable
+        mark_ready: true
+        install_type: fullstack
+        scc:
+          instance_id: 1c7d5f78-9262-44c3-b779-b28fe4d88c37
```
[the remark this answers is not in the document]
Khuzaima05: done

ibm_catalog.json
```diff
@@ -0,0 +1,533 @@
...
+      ],
+      "keywords": [
+        "vpc",
+        "terraform",
+        "IaC",
+        "infrastructure as code",
+        "solution"
+      ],
+      "short_description": "Provisions a VPC network on IBM Cloud",
+      "long_description": "The VPC deployable architecture deploys a Virtual Private Cloud (VPC) infrastructure without any compute resources, such as Virtual Server Instances (VSI) or Red Hat OpenShift clusters.",
```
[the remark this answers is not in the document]
Khuzaima05: Done

ibm_catalog.json
```diff
@@ -0,0 +1,533 @@
...
+        "solution"
+      ],
+      "keywords": [
+        "vpc",
+        "terraform",
+        "IaC",
+        "infrastructure as code",
+        "solution"
+      ],
+      "short_description": "Provisions a VPC network on IBM Cloud",
```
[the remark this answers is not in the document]
Khuzaima05: Done

.catalog-onboard-pipeline.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: v1
+offerings:
+  - name: deploy-arch-ibm-vpc
+    kind: solution
+    catalog_id: f64499c8-eb50-4985-bf91-29f9e605a433
+    offering_id: ae881fc1-273f-472e-8437-99099a6a753b
```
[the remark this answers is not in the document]
Khuzaima05: updated id

ibm_catalog.json
```diff
@@ -0,0 +1,533 @@
...
+              }
+            ],
+            "diagrams": [
+              {
+                "diagram": {
+                  "caption": "VPC for IBM Cloud.",
+                  "url": "https://raw.githubusercontent.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/main/reference-architecture/deployable-architecture-vpc.svg",
+                  "type": "image/svg+xml"
+                },
+                "description": "This architecture creates a VPC."
```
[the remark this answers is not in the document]
Khuzaima05: updated

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,694 @@
...
+variable "flow_logs_cos_bucket_default_retention_days" {
+  description = "The number of days that an object can remain unmodified in the flow logs cloud object storage bucket."
+  type        = number
+  default     = 90
+}
+
+variable "flow_logs_cos_bucket_maximum_retention_days" {
+  description = "The maximum number of days that an object can be kept unmodified in the flow logs cloud object storage."
+  type        = number
+  default     = 350
```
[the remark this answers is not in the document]
maheshwarishikha: For retention...thats the value we keep it. I'll check docs more on it. We use these values everytime.
https://ibm-cloudplatform.slack.com/archives/C0117873D0S/p1716805861637559

solutions/fully-configurable/variables.tf
```diff
@@ -0,0 +1,694 @@
...
+variable "flow_logs_cos_bucket_archive_days" {
+  description = "The number of days before the `archive_type` rule action takes effect for the flow logs cloud object storage bucket."
+  type        = number
+  default     = 90
+}
+
+variable "flow_logs_cos_bucket_archive_type" {
+  description = "The storage class or archive type you want the object to transition to in the flow logs cloud object storage bucket."
+  type        = string
+  default     = "Glacier"
```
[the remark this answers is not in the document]
Khuzaima05: Done

tests/pr_test.go
```diff
@@ -201,3 +203,93 @@ func TestRunVpcWithDnsExample(t *testing.T) {
...
+		TarIncludePatterns: []string{
+			"*.tf",
+			"dynamic_values/*.tf",
+			"dynamic_values/config_modules/*/*.tf",
+			fullyConfigFlavorDir + "/*.tf",
+		},
+		TemplateFolder:         fullyConfigFlavorDir,
+		Tags:                   []string{"vpc-da-test"},
+		DeleteWorkspaceOnFail:  false,
+		WaitJobCompleteMinutes: 60,
```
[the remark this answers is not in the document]
Khuzaima05: increased to 120

.catalog-onboard-pipeline.yaml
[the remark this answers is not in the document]
maheshwarishikha: updated

ibm_catalog.json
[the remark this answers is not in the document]
maheshwarishikha: We may be doing the same for other DAs going forward. 
Refer - https://github.com/terraform-ibm-modules/terraform-ibm-scc/blob/main/ibm_catalog.json#L386

ibm_catalog.json
```diff
@@ -0,0 +1,543 @@
...
+      ],
+      "keywords": [
+        "vpc",
+        "terraform",
+        "IaC",
+        "infrastructure as code",
+        "solution"
+      ],
+      "short_description": "Automates VPC deployment on IBM Cloud, offering full configurability and flexibility for diverse workloads.",
+      "long_description": "The VPC deployable architecture deploys a Virtual Private Cloud (VPC) infrastructure without any compute resources, such as Virtual Server Instances (VSI) or Red Hat OpenShift clusters.This is an experimental tile and not suitable for production workloads. Stay here if you want to try an experimental version with the Optional and swappable components capability.",
```
rajatagarwal-ibm, line 23: can you please add the URLs pointing to the optional and swappable components as well?

ibm_catalog.json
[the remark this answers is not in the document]
rajatagarwal-ibm: you're right, my bad.

rajatagarwal-ibm (changes requested):

ibm_catalog.json
maheshwarishikha: added

rajatagarwal-ibm (dismissed):
LGTM

ocofaigh (dismissed):

ibm_catalog.json
```diff
@@ -0,0 +1,543 @@
+{
+  "products": [
+    {
+      "name": "deploy-arch-ibm-vpc",
+      "label": "[Add-ons Beta] Cloud automation for VPC",
+      "product_kind": "solution",
+      "tags": [
+        "network_vpc",
```
ocofaigh, line 8: @maheshwarishikha This is a subcategory tag. We are missing a main category tag. I'm going to add the `network` tag or it will prevent us from publishing. FYI you can see all tags by running `ibmcloud catalog filter options --all`

ibm_catalog.json
```diff
@@ -0,0 +1,543 @@
...
+        "ibm_created",
+        "target_terraform",
+        "terraform",
+        "support_ibm",
```
ocofaigh, line 12: @maheshwarishikha I think we need to remove this and add the `ibm_beta` tag. We are not supporting this tile via IBM Cloud - its only experimental beta

ocofaigh (approved):

Issue #989, opened 2025-05-13
VPC DA is missing a test with flow logs and bucket encryption enabled


ocofaigh:
We have a test gap - there is no test that tests setting `kms_encryption_enabled_bucket` to true, or `enable_vpc_flow_logs` to true

ocofaigh:
@Vipin654 I merged https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/pull/992 but I think we need a follow up PR to:

- Remove `TestRunUpgradeDefaultExample` - we already have an upgrade test for the DA
- Update the DA upgrade test `TestRunUpgradeFullyConfigurable` to also enable flow logs and kms encryption. Otherwise we might miss breaking changes related to those features

iamar7:
> We have a test gap - there is no test that tests setting kms_encryption_enabled_bucket to true, or enable_vpc_flow_logs to true

Both the settings are being testing in the `TestFullyConfigurableWithFlowLogs` -  https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/blob/d525ef5c3d94e9e61ea729303d1bd8811b4b457a/tests/pr_test.go#L315

https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/blob/d525ef5c3d94e9e61ea729303d1bd8811b4b457a/tests/pr_test.go#L317

iamar7:
> * Remove `TestRunUpgradeDefaultExample` - we already have an upgrade test for the DA
> * Update the DA upgrade test `TestRunUpgradeFullyConfigurable` to also enable flow logs and kms encryption. Otherwise we might miss breaking changes related to those features

This was covered in https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/pull/994

PR #992, opened 2025-05-14
test: added flow logs test case

Vipin654:
### Description
test: added flow logs test case
[Issue](https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/issues/989)

### Release required?

- [x] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers

### DIFF
--- a/tests/pr_test.go
+++ b/tests/pr_test.go
@@ -248,6 +248,53 @@ func TestFullyConfigurable(t *testing.T) {
 	assert.Nil(t, err, "This should not have errored")
 }
 
+func TestFullyConfigurableWithFlowLogs(t *testing.T) {
+	t.Parallel()
+
+	// Verify ibmcloud_api_key variable is set
+	checkVariable := "TF_VAR_ibmcloud_api_key"
+	val, present := os.LookupEnv(checkVariable)
+	require.True(t, present, checkVariable+" environment variable not set")
+	require.NotEqual(t, "", val, checkVariable+" environment variable is empty")
+
+	// Programmatically determine region to use based on availability
+	region, _ := testhelper.GetBestVpcRegion(val, "../common-dev-assets/common-go-assets/cloudinfo-region-vpc-gen2-prefs.yaml", "eu-de")
+
+	prefix := "vpc-da-fl"
+
+	options := testschematic.TestSchematicOptionsDefault(&testschematic.TestSchematicOptions{
+		Testing: t,
+		Region:  region,
+		Prefix:  prefix,
+		TarIncludePatterns: []string{
+			"*.tf",
+			"dynamic_values/*.tf",
+			"dynamic_values/config_modules/*/*.tf",
+			fullyConfigFlavorDir + "/*.tf",
+		},
+		TemplateFolder:         fullyConfigFlavorDir,
+		Tags:                   []string{"vpc-da-test"},
+		DeleteWorkspaceOnFail:  false,
+		WaitJobCompleteMinutes: 120,
+	})
+
+	options.TerraformVars = []testschematic.TestSchematicTerraformVar{
+		{Name: "ibmcloud_api_key", Value: options.RequiredEnvironmentVars["TF_VAR_ibmcloud_api_key"], DataType: "string", Secure: true},
+		{Name: "existing_resource_group_name", Value: resourceGroup, DataType: "string"},
+		{Name: "region", Value: options.Region, DataType: "string"},
+		{Name: "resource_tags", Value: options.Tags, DataType: "list(string)"},
+		{Name: "access_tags", Value: permanentResources["accessTags"], DataType: "list(string)"},
+		{Name: "prefix", Value: options.Prefix, DataType: "string"},
+		{Name: "enable_vpc_flow_logs", Value: "true", DataType: "bool"},
+		{Name: "existing_cos_instance_crn", Value: permanentResources["general_test_storage_cos_instance_crn"], DataType: "string"},
+		{Name: "kms_encryption_enabled_bucket", Value: "true", DataType: "bool"},
+		{Name: "existing_kms_instance_crn", Value: permanentResources["hpcs_south_crn"], DataType: "string"},
+	}
+
+	err := options.RunSchematicTest()
+	assert.Nil(t, err, "This should not have errored")
+}
+
 // Test the upgrade of fully-configurable DA with defaults
 func TestRunUpgradeFullyConfigurable(t *testing.T) {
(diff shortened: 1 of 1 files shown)


rajatagarwal-ibm:
@Vipin654, no release is required for changes in the tests.

tests/pr_test.go
```diff
@@ -248,6 +248,53 @@ func TestFullyConfigurable(t *testing.T) {
...
+	// Verify ibmcloud_api_key variable is set
+	checkVariable := "TF_VAR_ibmcloud_api_key"
+	val, present := os.LookupEnv(checkVariable)
+	require.True(t, present, checkVariable+" environment variable not set")
+	require.NotEqual(t, "", val, checkVariable+" environment variable is empty")
+
+	// Programmatically determine region to use based on availability
+	region, _ := testhelper.GetBestVpcRegion(val, "../common-dev-assets/common-go-assets/cloudinfo-region-vpc-gen2-prefs.yaml", "eu-de")
+
+	prefix := "vpc-da-fl"
```
rajatagarwal-ibm, line 263: Could you generate a random prefix here? Here's the format 
```
prefix := fmt.Sprintf("existing-vpc-test-%s", strings.ToLower(random.UniqueId()))
```

Vipin654:
updated the PR @rajatagarwal-ibm

tests/pr_test.go
```diff
@@ -248,6 +248,53 @@ func TestFullyConfigurable(t *testing.T) {
...
+
+	options.TerraformVars = []testschematic.TestSchematicTerraformVar{
+		{Name: "ibmcloud_api_key", Value: options.RequiredEnvironmentVars["TF_VAR_ibmcloud_api_key"], DataType: "string", Secure: true},
+		{Name: "existing_resource_group_name", Value: resourceGroup, DataType: "string"},
+		{Name: "region", Value: options.Region, DataType: "string"},
+		{Name: "resource_tags", Value: options.Tags, DataType: "list(string)"},
+		{Name: "access_tags", Value: permanentResources["accessTags"], DataType: "list(string)"},
+		{Name: "prefix", Value: options.Prefix, DataType: "string"},
+		{Name: "enable_vpc_flow_logs", Value: "true", DataType: "bool"},
+		{Name: "existing_cos_instance_crn", Value: permanentResources["general_test_storage_cos_instance_crn"], DataType: "string"},
```
rajatagarwal-ibm, line 289: Since `enable_vpc_flow_logs` and `kms_encryption_enabled_bucket` are both set to `true`, a new authorisation policy between COS and KMS is created every time this test runs. However, because you're using permanent instances for both COS and KMS, running the tests back-to-back/parallel (without waiting for the previous auth policy to be destroyed) can lead to conflicts during policy creation.

To avoid this issue, I recommend modifying the test to use existing resources and ensure that at least one of the services—either COS or KMS—is recreated for each test run. This will help prevent policy duplication conflicts.

rajatagarwal-ibm (changes requested):

tests/existing-resources/main.tf
```diff
@@ -27,3 +27,9 @@ resource "ibm_is_subnet" "testacc_subnet" {
   total_ipv4_address_count = 256
   resource_group           = module.resource_group.resource_group_id
 }
+
+resource "ibm_resource_instance" "cos_instance" {
+  name              = var.cos_instance_name
+  resource_group_id = module.resource_group.resource_group_id
+  service           = "cloud-object-storage"
+}
```
ocofaigh, line 35: No need to create a COS instance, there is a permanent COS instance we can use for tests -> https://github.com/terraform-ibm-modules/common-dev-assets/blob/9fdee86807c74a35061fc3327c2a01215897cb7f/common-go-assets/common-permanent-resources.yaml#L174

tests/pr_test.go
Vipin654: addressed.

tests/pr_test.go
```diff
@@ -170,6 +170,43 @@ func TestRunExistingVPCExample(t *testing.T) {
 		output, err := options.RunTestConsistency()
 		assert.Nil(t, err, "This should not have errored")
 		assert.NotNil(t, output, "Expected some output")
+
+		// test for flow logs with encrypted bucket, new vpc will be created but existing cos instance will be used
```
ocofaigh, line 174: We should be using sub tests for this scenario - see https://github.ibm.com/GoldenEye/issues/issues/12430

However you don't need to create any resources for this test - you can use existing HPCS and COS instance and create a brand new test for this use case

ocofaigh (changes requested):
see comments

tests/pr_test.go
ocofaigh: This is not an going to be an issue - the s2s auth policy is scoped to the KMS key, and the DA will create a new KMS key so it will never clash

rajatagarwal-ibm:
@Vipin654, please revert the changes - the policies are bound to key and buckets instead of instances, so there should not be any conflicts.

rajatagarwal-ibm (approved):

ocofaigh (changes requested):
1 minor comment

tests/pr_test.go
```diff
@@ -248,6 +248,53 @@ func TestFullyConfigurable(t *testing.T) {
...
+	t.Parallel()
+
+	// Verify ibmcloud_api_key variable is set
+	checkVariable := "TF_VAR_ibmcloud_api_key"
+	val, present := os.LookupEnv(checkVariable)
+	require.True(t, present, checkVariable+" environment variable not set")
+	require.NotEqual(t, "", val, checkVariable+" environment variable is empty")
+
+	// Programmatically determine region to use based on availability
+	region, _ := testhelper.GetBestVpcRegion(val, "../common-dev-assets/common-go-assets/cloudinfo-region-vpc-gen2-prefs.yaml", "eu-de")
```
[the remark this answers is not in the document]
Vipin654: @ocofaigh i removed this line and Region: region while creating options.. I hope it is fine.. only difference here is that default region is eu-de but in testwrapper it would be us-south

ocofaigh (approved):

ocofaigh:
@Vipin654 Thanks for adding this - I think we have some more actions for these tests, please see https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/issues/989#issuecomment-2880985106

PR #1094, opened 2025-12-08
chore: add config_constraints to code_editor variables in ibm_catalog.json

piyush117:


shemau (changes requested):
Please include common-dev-assets, otherwise the changes are not validated in the pipeline.

PR #1098, opened 2025-12-10
fix: update subnet naming to resource_name to allow safe prefix updates

Khuzaima05:
### Description

1. Updates subnet and address-prefix resource naming to use a stable resource_name instead of names derived from the VPC prefix.
2. This prevents unintended resource destruction/recreation when the prefix value is changed.
3.  Ensures Terraform state remains consistent and safe for prefix updates on existing VPCs.
4.  Includes migration support to map existing state to the new naming scheme.

issue: https://github.ibm.com/GoldenEye/issues/issues/16771

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [x] Major release (`X.x.x`)

##### Release notes content

This release introduces resource-based subnet naming, enabling safe prefix changes without forcing subnet recreation.

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


Khuzaima05:
Skipping the upgrade test because this change updates the Terraform keys used for subnets and address prefixes . it will not find the existing resources. Upgrade will fail.

Screenshot 2026-01-20 at 10 25 51 AM

Khuzaima05:
> @Khuzaima05 The scripts you added are for module consumers, but what about the DA? We have several DAs that use this module. We need to full understand the impact here. DA users are not going to know about the scripts you added. Plus we have several other external consumes like PowerVS, HPC etc etc. We need to find a clean way to do this. In the past we had to use a pre-deploy ansible script for DAs. but its a lot of effort. I want to fully understand why we are making this breaking change after many years of using this code. Please add to the deep dive here so we can discuss.

@ocofaigh , sure I will add this topic in tomorrow’s deep dive.

PR #1134, opened 2026-02-16
fix: replace deprecated tcp/udp/icmp SG rule blocks

piyush117:
### Description

- Replace deprecated nested tcp/udp/icmp protocol blocks in ibm_is_security_group_rule and ibm_is_network_acl resources with flat top-level attributes: protocol, port_min, port_max, source_port_min, source_port_max, type, and code.
- Update all variable schemas (security_group, network_acls, load_balancers SG rules, etc..) across the module, submodules, solutions, and examples to use the same flat structure — removing the tcp, udp, and icmp optional block fields.
- Remove the intermediate translation locals that previously mapped block-style inputs to provider fields.

The IBM provider deprecated nested tcp/udp/icmp blocks in favour of flat protocol attributes. This PR fully adopts the flat approach end-to-end — both the resource definitions and the module's public variable API — eliminating deprecation warnings.

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [x] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


piyush117:
Old behaviour,
Screenshot 2026-02-16 at 12 05 53 PM
New behaviour,
Screenshot 2026-02-16 at 12 06 23 PM

default_security_group.tf
```diff
@@ -18,29 +18,27 @@ resource "ibm_is_security_group_rule" "default_vpc_rule" {
...
-      port_min = each.value.tcp.port_min
-      port_max = each.value.tcp.port_max
-    }
-  }
+  protocol = (
+    each.value.tcp != null ? "tcp" :
+    each.value.udp != null ? "udp" :
+    each.value.icmp != null ? "icmp" :
+    null
+  )
```
jor2: Set this to `"all"` instead of `null`. The provider requires an explicit protocol value, and an allow-all rule needs `"all"` here, not null.

jor2 (commented):
Clean migration to flat attributes. One issue: `protocol = null` when none of tcp/udp/icmp is set - should be `"all"`.

piyush117:
Upgrade tests were showing update in place,
tested it locally and verified that only the protocol change is happening.
Screenshot 2026-05-06 at 1 52 24 PM

Screenshot 2026-05-06 at 1 37 44 PM

Screenshot 2026-05-06 at 1 52 48 PM

Now since the protocol is changing from `icmp_tcp_udp` to `any`, made a small change i.e. if the action is `deny` it will pick `any` and for `allow` it will pick `icmp_tcp_udp`

Aashiq-J (commented):
On second thought, we should revise the variable structure for both `network_acl` and the security group. The current structure only supports `tcp`, `udp`, and `icmp`, even though the UI allows rules to be created for many other protocol types.

Updating the structure will make the module more flexible and future-proof for any additional protocols that may be introduced later.

variables.tf
```diff
@@ -512,18 +478,11 @@ variable "security_group_rules" {
...
-        # Count how many protocols are specified (non-null)
-        # Return false if more than one protocol is specified
-        false if length([
-          for protocol in [rule.tcp, rule.udp, rule.icmp] :
-          protocol if protocol != null
-        ]) > 1
-      ])
-    )) == 0
+    error_message = "Security group rule protocol must be one of: `tcp`, `udp`, `icmp`, `icmp_tcp_udp`."
+    condition = (var.security_group_rules == null || length(var.security_group_rules) == 0) ? true : alltrue([
```
Aashiq-J, line 483: This validation is no longer valid, check the UI/docs there are multiple new options of protocol to choose from apart from these 4.

variables.tf
```diff
@@ -461,24 +440,11 @@ variable "security_group_rules" {
...
-      icmp = optional(
-        object({
-          type = optional(number)
-          code = optional(number)
-        })
-      )
+      protocol   = optional(string)
+      port_min   = optional(number)
+      port_max   = optional(number)
+      type       = optional(number)
```
Aashiq-J, line 447: type and code are icmp related parameter we can have a validation for that.

variables.tf
```diff
@@ -297,17 +282,11 @@ variable "network_acls" {
...
       flatten([
-        # Check through rules
         for rule in flatten([var.network_acls[*].rules]) :
-        # Count how many protocols are specified (non-null)
-        # Return false if more than one protocol is specified
-        false if length([
-          for protocol in [rule.tcp, rule.udp, rule.icmp] :
-          protocol if protocol != null
-        ]) > 1
+        false if rule.protocol != null && !contains(["tcp", "udp", "icmp", "icmp_tcp_udp"], rule.protocol)
```
Aashiq-J: This validation is no longer valid, check the UI/docs there are multiple new options of protocol to choose from apart from these 4.

Aashiq-J (commented):
LGTM

Aashiq-J (approved):

